PR #23, opened 2023-01-13
first pass at CI

dignifiedquire:
Ref n0-computer/iroh#771


Arqu (approved):
Great for now. As time permits, I'll look into moving it to our own boxes.

PR #24, opened 2023-01-13
feat: implement custom version of libp2p-tls

dignifiedquire:
- only ed25519 keys are supported
- no support for PeerIds that are based on protobufs
- remove support for RSA on the certificates

### DIFF
--- a/src/tls/certificate.rs
+++ b/src/tls/certificate.rs
@@ -1,0 +1,383 @@
+//! X.509 certificate handling.
+//!
+//! This module handles generation, signing, and verification of certificates.
+
+use der::{asn1::OctetStringRef, Decode, Encode, Sequence};
+use x509_parser::prelude::*;
+
+use super::{Keypair, PeerId, PublicKey, Signature};
+
+/// The libp2p Public Key Extension is a X.509 extension
+/// with the Object Identier 1.3.6.1.4.1.53594.1.1,
+/// allocated by IANA to the libp2p project at Protocol Labs.
+const P2P_EXT_OID: [u64; 9] = [1, 3, 6, 1, 4, 1, 53594, 1, 1];
+
+/// The peer signs the concatenation of the string `libp2p-tls-handshake:`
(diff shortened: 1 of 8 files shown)


src/tls/certificate.rs
```diff
@@ -0,0 +1,383 @@
...
+
+use der::{asn1::OctetStringRef, Decode, Encode, Sequence};
+use x509_parser::prelude::*;
+
+use super::{Keypair, PeerId, PublicKey, Signature};
+
+/// The libp2p Public Key Extension is a X.509 extension
+/// with the Object Identier 1.3.6.1.4.1.53594.1.1,
+/// allocated by IANA to the libp2p project at Protocol Labs.
+const P2P_EXT_OID: [u64; 9] = [1, 3, 6, 1, 4, 1, 53594, 1, 1];
```
b5, line 13: can we get a TODO to not use this object identifier? It's tied to the protobuf you're trying to remove

src/tls/certificate.rs
```diff
@@ -0,0 +1,383 @@
...
+        msg.extend(P2P_SIGNING_PREFIX);
+        msg.extend(certificate_keypair.public_key_der());
+
+        identity_keypair.sign(&msg)
+    };
+
+    let public_key = identity_keypair.public();
+    let signature = signature.to_bytes();
+    let key = SignedKey {
+        public_key: OctetStringRef::new(&public_key.as_bytes()[..]).unwrap(),
```
b5, line 182: IIUC this is the crucial change: dropping the keytype enum & protobuf serialization in favour of just the ED publickey directly, right?

b5 (approved):
ran tests locally, am satisfied that things still work. The comments / spec side of this is a mess, but I vote we clean that up later

src/tls/certificate.rs
dignifiedquire: we don't want to allocate our own at the IANA for now, so for now I don't think we should change this

PR #31, opened 2023-01-19
feat: connect based on PeerIds

dignifiedquire:
This enforces passing the public key, to ensure the identity being connected to, is the one we expect.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -11,28 +11,30 @@ use tokio::io::{AsyncReadExt, AsyncWrite, AsyncWriteExt};
 use tracing::debug;
 
 use crate::protocol::{read_lp_data, write_lp, Handshake, Request, Res, Response};
-use crate::tls::{self, Keypair};
+use crate::tls::{self, Keypair, PeerId};
 
 const MAX_DATA_SIZE: usize = 1024 * 1024 * 1024;
 
 #[derive(Clone, Debug)]
 pub struct Options {
     pub addr: SocketAddr,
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -31,6 +32,8 @@ enum Commands {
     #[clap(about = "Fetch the data from the hash")]
     Client {
         hash: bao::Hash,
+        #[clap(long)]
```
ramfox, line 35: can we get a comment here specifying it's the `PeerId` of the server
`/// The server's peer identifier`
or something?

Just to make it clear that it's the server's peer id and not the clients.

src/client.rs
```diff
@@ -11,28 +11,30 @@ use tokio::io::{AsyncReadExt, AsyncWrite, AsyncWriteExt};
...
         Options {
             addr: "127.0.0.1:4433".parse().unwrap(),
+            peer_id: None,
         }
     }
 }
 
 /// Setup a QUIC connection to the provided server address
-async fn setup(server_addr: SocketAddr) -> Result<(Client, Connection)> {
+async fn setup(server_addr: SocketAddr, peer_id: Option<PeerId>) -> Result<(Client, Connection)> {
```
ramfox:
```suggestion
async fn setup(opts: Options) -> Result<(Client, Connection)> {
```
?
We pass all the options here and never refer to them again in the `run` function.

ramfox (approved):
Looks good!
two comments, but could be merged as is

PR #37, opened 2023-01-20
feat: rework client api to emit content

dignifiedquire:
This prepares the api to allow for multiple results to be emitted

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -68,11 +68,11 @@ pub enum Event {
         size: usize,
     },
     /// Content is being received.
-    Received {
+    Receiving {
         /// The hash of the content we received.
         hash: bao::Hash,
         /// The actual data we are receiving.
-        data: Box<dyn AsyncRead + Unpin + Sync + Send + 'static>,
+        reader: Box<dyn AsyncRead + Unpin + Sync + Send + 'static>,
     },
     /// The transfer is done.
     Done(Stats),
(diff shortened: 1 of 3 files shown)


src/client.rs
```diff
@@ -58,17 +58,27 @@ pub struct Stats {
...
+    /// The connection to the server was established.
     Connected,
-    Requested { size: usize },
+    /// The server has the content.
+    Requested {
+        /// The size of the requested content.
+        size: usize,
+    },
+    /// Content is being received.
+    Received {
```
flub: Maybe this should be called "Receiving"?

src/client.rs
```diff
@@ -58,17 +58,27 @@ pub struct Stats {
...
+        /// The hash of the content we received.
+        hash: bao::Hash,
+        /// The actual data we are receiving.
+        data: Box<dyn AsyncRead + Unpin + Sync + Send + 'static>,
```
flub: nit: maybe call this `reader` rather than `data`?  It also affects the variables you end up using in the code and I think they would be clearer as `reader`.

flub (approved):

src/client.rs
dignifiedquire: fixed

src/client.rs
dignifiedquire: fixed

PR #38, opened 2023-01-20
feat: add STDIN and STDOUT support

### LINKED ISSUE #35: feat: outboard storage
We currently hold the data in memory. We don't want to duplicate data unnecessarily. We need to implement an outboard storage method.

### LINKED ISSUE #752: feat: outboard storage
We currently hold the data in memory. We don't want to duplicate data unnecessarily. We need to implement an outboard storage method.

dignifiedquire:
Also avoid keeping files data in memory, after calculating the outboard data.

Based on n0-computer/sendme#37 
Fixes n0-computer/iroh#752 

Screenshot 2023-01-20 at 14 36 02

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -6,7 +6,7 @@ use anyhow::{anyhow, bail, ensure, Result};
 use bytes::{Bytes, BytesMut};
 use s2n_quic::stream::BidirectionalStream;
 use s2n_quic::Server;
-use tokio::io::{AsyncReadExt, AsyncWrite, AsyncWriteExt};
+use tokio::io::AsyncWrite;
 use tracing::{debug, warn};
 
 use crate::protocol::{read_lp, write_lp, Handshake, Request, Res, Response, VERSION};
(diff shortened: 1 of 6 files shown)


src/server.rs
```diff
@@ -142,33 +156,87 @@ async fn handle_stream(
...
+                    Data::File {
+                        outboard: Bytes::from(outboard),
+                        path,
+                        size: data.len(),
+                    },
+                );
+            }
+            DataSource::Stdin(mut stdin) => {
+                let mut data = BytesMut::new();
+                while stdin.read_buf(&mut data).await? > 0 {}
```
rklaehn: Would it really be so bad to save this into a temp file that gets deleted on process exit?

Seems like the unix thing to do, and it would work even if you do `head -c 100_000_000_000 /dev/urandom | sendme --server`. Also, you then got only one kind of data in your db - file paths.

dignifiedquire: > Seems like the unix thing to do

Do you have an example of what tools do this?

flub:
Heads up, I changed the base of this PR to make the diff more readable

src/server.rs
flub: I have to say I do like this temp file idea... rather much

Though I don't think it means you only have to store one thing in the db:

- You'd want to have this temp file(s) only by file descriptor and not by path, so that they don't even have to be on the dirents.  This also helps with being robust.
- Now you have a portability problem: being a good unix citizen made windows harder.

flub: Since we're supposed to be a library I wonder if you could store something more generic instead of the full data.  What if you stored a boxed reader?  That way it's up to the implementer how they back this reader: by bytes in memory, by tempfiles, by calls to an external database process, ...

dignifiedquire:
Closing in favor of https://github.com/n0-computer/sendme/pull/39

src/server.rs
dignifiedquire: In #39 this now uses a tempfile, not as fancy as you described, but better than just in memory, I agree

rklaehn: > > Seems like the unix thing to do
> 
> Do you have an example of what tools do this?

Sort. If you pipe a large file to sort it will create temp files.

PR #39, opened 2023-01-20
feat: add STDIN and STDOUT support

dignifiedquire:
also avoid keeping files data in memory, after calculating the outboard data

Rebase issues.. Replaces #38

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,10 +20,10 @@ struct Cli {
 #[derive(Subcommand, Debug, Clone)]
 #[allow(clippy::large_enum_variant)]
 enum Commands {
-    /// Serve the data from the given path
+    /// Serve the data from the given path. If none is specified reads from STDIN.
     #[clap(about = "Serve the data from the given path")]
     Server {
-        paths: Vec<PathBuf>,
+        path: Option<PathBuf>,
         #[clap(long, short)]
         /// Optional port, defaults to 127.0.01:4433.
         addr: Option<SocketAddr>,
@@ -36,10 +36,9 @@ enum Commands {
         /// PeerId of the server.
         peer_id: PeerId,
         #[clap(long, short)]
-        /// Option address of the server, defaults to 127.0.0.1:4433.
+        /// Optional address of the server, defaults to 127.0.0.1:4433.
         addr: Option<SocketAddr>,
-        #[clap(long, short)]
-        /// Option path to save the file, defaults to using the hash as the name.
+        /// Optional path to save the file. If none is specified writes the data to STDOUT.
         out: Option<PathBuf>,
     },
 }
@@ -69,9 +68,6 @@ async fn main() -> Result<()> {
                 opts.addr = addr;
             }
 
-            // Write file out
-            let outpath = out.unwrap_or_else(|| hash.to_string().into());
-
             println!("{} Connecting ...", style("[1/3]").bold().dim());
             let pb = ProgressBar::hidden();
             let stream = client::run(hash, opts);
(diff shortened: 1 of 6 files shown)


src/server.rs
```diff
@@ -159,33 +157,67 @@ async fn handle_stream(
...
+
+async fn write_response<W: AsyncWrite + Unpin>(
+    mut writer: W,
+    buffer: &mut BytesMut,
+    id: u64,
+    res: Res<'_>,
+) -> Result<()> {
+    let response = Response { id, data: res };
+
+    if buffer.len() < 20 + response.data.len() {
```
rklaehn, line 214: Where does this 20 come from? Is it the postcard overhead over the raw data?

src/main.rs
```diff
@@ -20,10 +20,10 @@ struct Cli {
 #[derive(Subcommand, Debug, Clone)]
 #[allow(clippy::large_enum_variant)]
 enum Commands {
-    /// Serve the data from the given path
+    /// Serve the data from the given path. If none is specified reads from STDIN.
     #[clap(about = "Serve the data from the given path")]
     Server {
-        paths: Vec<PathBuf>,
+        path: Option<PathBuf>,
```
rklaehn, line 26: So we can only serve a single path now?

src/main.rs
```diff
@@ -111,16 +113,33 @@ async fn main() -> Result<()> {
...
+            let sources = if let Some(path) = path {
+                vec![server::DataSource::File(path)]
+            } else {
+                // Store STDIN content into a temporary file
+                let (file, path) = tempfile::NamedTempFile::new()?.into_parts();
+                let mut file = tokio::fs::File::from_std(file);
+                let path_buf = path.to_path_buf();
+                tmp_path = Some(path);
+                tokio::io::copy(&mut tokio::io::stdin(), &mut file).await?;
+                vec![server::DataSource::File(path_buf)]
```
rklaehn, line 128: works for me. sort does some optimizations to only create temp files when the input is large, but I think this is fine for testing, and the typical sendme use case will be files that are larger than you want in memory on a phone.

rklaehn (approved):

src/server.rs
```diff
@@ -159,33 +157,67 @@ async fn handle_stream(
...
 
-#[derive(Clone)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct Data {
-    /// outboard data from bo
+    /// Outboard data from bao.
     outboard: Bytes,
-    /// actual data
-    data: Bytes,
+    /// Path to the data.
```
flub:
```suggestion
    /// Path to the original data, which must not change while in use.
```

src/server.rs
```diff
@@ -159,33 +157,67 @@ async fn handle_stream(
...
+    path: PathBuf,
+    /// Size of the data.
```
flub:
```suggestion
    /// Size of the original data.
```

src/server.rs
```diff
@@ -159,33 +157,67 @@ async fn handle_stream(
...
+    buffer: &mut BytesMut,
+    id: u64,
+    res: Res<'_>,
+) -> Result<()> {
+    let response = Response { id, data: res };
+
+    if buffer.len() < 20 + response.data.len() {
+        buffer.resize(20 + response.data.len(), 0u8);
+    }
+    let used = postcard::to_slice(&response, buffer)?;
```
flub, line 217: We go to pains to use a tempfile for stdin but then here just load whatever entirely in memory?  I think this is fine for here, but I'm going to create an issue to address this and see if we can do this streaming (I can't immediately see how with postcard though...)

flub (approved):

src/server.rs
dignifiedquire: yeah, that‘s been here since the beginning, it’s my estimate of overhead of the struct overhead

src/server.rs
dignifiedquire: this is only the outboard data, the io::copy streams things from disk

src/main.rs
dignifiedquire: yes, mostly did this for making the api simpler, we can always add multiple paths back if we want to later

src/server.rs
flub: oh right indeed.  this is only the boa outboard data.  apologies

flub (approved):

PR #47, opened 2023-01-23
feat: Add authentication token to protocol handshake

flub:
This ensures that only clients which have the token can connect to the
server and retrieve the file.

# status

This needs a little cleanup but is basically ready.  Pushing this now
so it can already have some feedback, after I've been very delayed due
to entirely self-inflicted needless bugs that I introduced.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,3 +1,4 @@
+use std::fmt::Debug;
 use std::time::Duration;
 use std::{io::Read, net::SocketAddr, time::Instant};
 
@@ -10,7 +11,7 @@ use s2n_quic::{client::Connect, Client};
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWriteExt};
 use tracing::debug;
 
-use crate::protocol::{read_lp_data, write_lp, Handshake, Request, Res, Response};
+use crate::protocol::{read_lp_data, write_lp, AuthToken, Handshake, Request, Res, Response};
 use crate::tls::{self, Keypair, PeerId};
 
 const MAX_DATA_SIZE: usize = 1024 * 1024 * 1024;
@@ -52,6 +53,7 @@ async fn setup(opts: Options) -> Result<(Client, Connection)> {
 }
 
 /// Stats about the transfer.
+#[derive(Debug)]
 pub struct Stats {
     pub data_len: usize,
     pub elapsed: Duration,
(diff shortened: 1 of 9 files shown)


src/client.rs
```diff
@@ -78,7 +80,25 @@ pub enum Event {
     Done(Stats),
 }
 
-pub fn run(hash: bao::Hash, opts: Options) -> impl Stream<Item = Result<Event>> {
+impl Debug for Event {
```
dignifiedquire: 😭 we should look at some crates like derive-more if we can use that to avoid these manual debug impls, they are so easy to get out of date (not needed for this PR, more a general note)

src/lib.rs
```diff
@@ -30,16 +31,21 @@ mod tests {
         let addr = "127.0.0.1:4443".parse().unwrap();
         let mut server = server::Server::new(db);
         let peer_id = server.peer_id();
+        let token = server.auth_token();
 
         tokio::task::spawn(async move {
-            server.run(server::Options { addr }).await.unwrap();
+            let opts = server::Options {
+                addr,
+                // token: Some(token),
```
dignifiedquire: why is this commented out?

src/main.rs
```diff
@@ -67,12 +73,15 @@ async fn main() -> Result<()> {
...
+            let token =
+                AuthToken::from_hex(&token).context("Wrong format for authentication token")?;
 
             println!("{} Connecting ...", style("[1/3]").bold().dim());
             let pb = ProgressBar::hidden();
-            let stream = client::run(hash, opts);
+            let stream = client::run(hash, token, opts);
             tokio::pin!(stream);
             while let Some(event) = stream.next().await {
+                trace!("I HAZ EVENT from client: {:?}", event);
```
dignifiedquire: beautiful

src/server.rs
```diff
@@ -102,6 +114,7 @@ async fn handle_stream(
             VERSION,
             handshake.version
         );
+        ensure!(handshake.token == token, "AuthToken mismatch");
```
dignifiedquire: We should probably start thinking of sending an error when we abort the connection. Not sure.

src/lib.rs
flub: that's from an earlier version where the token was in the options.  but now there's a server struct it makes more sense there.  that's what i meant with "still needs to be tidied up"

dignifiedquire:
generally looks good :)

src/server.rs
flub: I was wondering this as well, I think we don't have to do this just yet though.  There's little value in knowing whether you're rejected because of the version of auth token right now.

Also using error control flow for rejecting connections is not very nice, as demonstrated by the warn log message right now.  But we can also improve this later.

dignifiedquire: 👍

flub:
PTAL

I think the `Options` should be turned into a `Builder` and not be passed to `Provider::run`.  But I also think it's best to leave that as a separate followup.

dignifiedquire (approved):
lgtm

PR #48, opened 2023-01-24
feat: transfer multiple files

### LINKED ISSUE #36: feat: transfer a `ListOfBlobs`
Allow transferring multiple files.

```
use bao::Hash;

struct ListOfBlobs {
		name: String,
		blobs: Vec<Blob>
}

struct Blob {
		name: String,
		data_hash: Hash,
}
```

& adjust protocol to expect multiple files

### LINKED ISSUE #751: feat: transfer a `ListOfBlobs`
Allow transferring multiple files.

```
use bao::Hash;

struct ListOfBlobs {
		name: String,
		blobs: Vec<Blob>
}

struct Blob {
		name: String,
		data_hash: Hash,
}
```

& adjust protocol to expect multiple files

ramfox:
closes n0-computer/iroh#751 

### Decisions I made that are up for debate

- The word `ListOfBlobs` was unpleasing to look at and to use for variable names. I changed it to `Collection`, which is a word that can maybe grow with us as `sendme` grows in capabilities. Anyway, can change it back, but I like the name.
- We don't "pre-check" the hashes to see if any are missing before starting a transfer. Trying to get to the actual "transfering" asap.

- As soon as a hash is `NotFound`, we end the transfer. We could continue trying to send other hashes in the list (if there are any).

- In the case of an error, we don't clean up any files that have already been transfered successfully. This seems like a reasonable expectation to me, but open to be disagreed with.

- In the case of an error, we don't report it to the remote, we just close the connection / stop the transfer.

- The `out` flag is now the expected directory you want to save the given file. We use the filename from the original path to name the file. In the case of stdin, this is the hash of the file.
- The `MAX_DATA_SIZE` I have as a limit for the entire `Collection` not individual files in the collection, so the `MAX_DATA_SIZE` is still the
  maximum we are willing to transfer. If this was not the intention lmk!

- The cli "transfering" progress bars are off, since the only information we have in main (in the current set up) is the TOTAL amount of data we are sending with the collection, not the sizes of the individual files. Open for us to pass on more information in the `Event`s but just didn't want to overburden this PR.

- There is currently a separate (mem) database for `Collection` data than the blobs themselves. It seems reasonable to me to instead store the `Collection` in a tempfile, and store them in the same database as the blobs. This would streamline our DB story and also streamline the protocol code a bit.

### General Request
- The buffering story in sendme is pretty hard to follow and debug currently.  
  Also, we probably do too much of it. Happy to take advice on this front!

### CLI usage
```
sendme server first_file second_file third_file
Available Data:
- a2d21fd875303f42cf43c78e778a2e5439407ea26687ba2eb98000ba8d6d85c3: 287264485 bytes
- 07ac3b46a681779978f808c4894f2d59552239203aa962bae91707a6307afdc9: 66659 bytes
- 06ca947e27c8f7ebf1e6ffc395f7f140c66dbb80ca8e05fda79ede8b7a06dba2: 1507262 bytes

collection:
- 3d647c0dd31ef5ae744f00675dd5c30efcee47a96308fed650dffc9256569e7c

Serving from c03094751d5848565ff15c49137ebfa0a98bb49e702ea7d13199dc00a13c688c

```

```
# transfer the whole collection
$ sendme client 06ca947e27c8f7ebf1e6ffc395f7f140c66dbb80ca8e05fda79ede8b7a06dba2 new_out_dir --peer-id c03094751d5848565ff15c49137ebfa0a98bb49e702ea7d13199dc00a13c688c

...

$ ls new_out_dir
first_file
second_file
third_file
```

```
# transfer a single file from the collection
$ sendme client 07ac3b46a681779978f808c4894f2d59552239203aa962bae91707a6307afdc9 new_out_dir --peer-id c03094751d5848565ff15c49137ebfa0a98bb49e702ea7d13199dc00a13c688c

...

$ ls new_out_dir
07ac3b46a681779978f808c4894f2d59552239203aa962bae91707a6307afdc9
```

### Rough protocol walkthrough

Server: waits for a request with a hash
- if the server has the hash in its collection database:
    -  it sends back a `Res::FoundCollection`, with the size of the encoded `Collection` data, the size of the actual amount of data we want to transfer, and the bao outboard data. Then it immediately sends the bao encoded data.
   - it decodes the Collection data, and iterates through each blob in the collection
        - if that blob hash can be found in the blob datasbase, it sends a `Res::Found`, with the size of the raw blob data & the bao outboard data, followed immediately with the bao encoded data
        - if it cannot be found, it sends back `Res::NotFound` and ends the transfer
- if the server has the hash in its blob database:
    - it sends back a `Res::Found`, with the size of the blob raw data, and the bao outboard data
- if the server cannot find the hash:
    -  it sends back `Res::NotFound` and ends the transfer

Client: requests a hash
- it waits for a response
- `Res::FoundCollection`
   - it reads the encoded `Collection` data, decodes it, and parses it into a `Collection`
   - it expects the blobs to come in in the order of the `Collection.blobs` vector field
   - for each blob:
       - it read off a `Res::Found` message, followed by the bao encoded data
       - it decodes the data & writes it to a duplexstream
       - the other end of the duplex stream is yielded in an `Event`, waiting to be read
    - if it gets to a `Res::NotFound` message, it ends the transfer
 - `Res::Found`
       - it read off a `Res::Found` message, followed by the bao encoded data
       - it decodes the data & writes it to a duplexstream
       - the other end of the duplex stream is yielded in an `Event`, waiting to be read
-  `Res::NotFound`
    - it ends the transfer


src/blobs.rs
```diff
@@ -0,0 +1,19 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub(crate) struct Collection {
+    /// The name of this collection
+    pub(crate) name: String,
+    /// Links to other blobs in this collection
```
flub:
```suggestion
    /// Links to the blobs in this collection
```

src/blobs.rs
```diff
@@ -0,0 +1,19 @@
...
+    /// The total size of the raw_data referred to by all links
+    pub(crate) raw_size: usize,
+}
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub(crate) struct Blob {
+    /// The name of this blob of data
+    pub(crate) name: String,
+    /// The hash of the blob of data
+    pub(crate) hash: [u8; 32],
```
flub: I think we've been using `bao::Hash` directly for hashes.  I don't see why we shouldn't do that?  In any case I'm a big fan of newtypes vs an array of bytes.

src/client.rs
```diff
@@ -73,11 +76,30 @@ pub enum Event {
...
+                write!(
+                    f,
+                    "Event::Receiving {{ hash: {hash}, reader: Box<AsyncReader>, name: {:#?} }}",
+                    name
+                )
+            }
+            Event::Done(s) => write!(f, "Event::Done({:#?})", s),
+        }
+    }
+}
```
flub: oh, your Debug impl of this is much nicer than mine!

src/client.rs
```diff
@@ -203,3 +233,92 @@ pub fn run(hash: bao::Hash, opts: Options) -> impl Stream<Item = Result<Event>>
...
+/// reads data into a passed in buffer, returning a buffer of length `size`, and clearing
+/// the original buffer of the already read data
+async fn read_data<R: AsyncRead + Unpin>(
+    size: usize,
+    mut reader: R,
+    mut buffer: &mut BytesMut,
+) -> Result<BytesMut> {
+    // TODO: avoid buffering
+    while buffer.len() < size {
+        reader.read_buf(&mut buffer).await?;
```
flub: needs a check for EOF from the reader i think

src/client.rs
```diff
@@ -203,3 +233,92 @@ pub fn run(hash: bao::Hash, opts: Options) -> impl Stream<Item = Result<Event>>
...
     }
 }
+
+/// reads data into a passed in buffer, returning a buffer of length `size`, and clearing
+/// the original buffer of the already read data
+async fn read_data<R: AsyncRead + Unpin>(
+    size: usize,
+    mut reader: R,
+    mut buffer: &mut BytesMut,
+) -> Result<BytesMut> {
```
flub: consider returing a `Bytes` instead here?  (you can call `BytesMut::freeze`)

src/client.rs
```diff
@@ -203,3 +233,92 @@ pub fn run(hash: bao::Hash, opts: Options) -> impl Stream<Item = Result<Event>>
...
+    // TODO: avoid buffering
+    while buffer.len() < size {
+        reader.read_buf(&mut buffer).await?;
+    }
+    if size > buffer.len() {
```
flub: with the current while loop this should not be possible?

src/client.rs
```diff
@@ -203,3 +233,92 @@ pub fn run(hash: bao::Hash, opts: Options) -> impl Stream<Item = Result<Event>>
...
+        bao::decode::Decoder::new_outboard(std::io::Cursor::new(&data[..]), &*outboard, &hash);
+
+    let mut buf = [0u8; 1024];
+    loop {
+        // TODO: avoid blocking
+        let read = decoder.read(&mut buf)?;
+        if read == 0 {
+            break;
+        }
+    }
```
flub: It took me ages to realise that was is being read here is exactly the same as what is already in `data`, with the sole purpose of verifying the hash.  I guess this will be nicer once we can have an async bao decoder.  In the mean time maybe add a comment saying "verifying the hash of `data`" or so?

src/client.rs
```diff
@@ -203,3 +233,92 @@ pub fn run(hash: bao::Hash, opts: Options) -> impl Stream<Item = Result<Event>>
...
+    let data = read_data(size, reader, buffer).await?;
+    // TODO: avoid copy
+    let outboard = outboard.to_vec();
+    let mut decoder =
+        bao::decode::Decoder::new_outboard(std::io::Cursor::new(&data[..]), &*outboard, &hash);
+
+    let mut buf = [0u8; 1024];
+    loop {
+        // TODO: avoid blocking
+        let read = decoder.read(&mut buf)?;
```
flub: What does the boa error message look like?  It might be worth adding a `.context("hash of Collection data does not match")?` or so

src/client.rs
```diff
@@ -203,3 +233,92 @@ pub fn run(hash: bao::Hash, opts: Options) -> impl Stream<Item = Result<Event>>
...
+
+async fn read_and_decode_blob_data<R: AsyncRead + Unpin>(
+    size: usize,
+    outboard: &[u8],
+    hash: bao::Hash,
+    name: Option<String>,
+    reader: R,
+    buffer: &mut BytesMut,
+) -> Result<(Event, tokio::task::JoinHandle<Result<()>>)> {
+    let data = read_data(size, reader, buffer).await?;
```
flub: do i understand correctly that this reads the entire blob into memory?  we should probably leave a comment so that this is easy to find in the future if so.

src/client.rs
```diff
@@ -203,3 +233,92 @@ pub fn run(hash: bao::Hash, opts: Options) -> impl Stream<Item = Result<Event>>
...
+
+    // TODO: avoid copy
+    let outboard = outboard.to_vec();
+    let t = tokio::task::spawn(async move {
+        let mut decoder =
+            bao::decode::Decoder::new_outboard(std::io::Cursor::new(&data[..]), &*outboard, &hash);
+
+        let mut buf = [0u8; 1024];
+        loop {
+            // TODO: avoid blocking
```
flub: where does this block right now?

src/main.rs
```diff
@@ -89,19 +89,21 @@ async fn main() -> Result<()> {
...
-                        ensure!(hash == new_hash, "invalid hash received");
+                        println!("  {}", style(format!("Receiving {}...", hash)).bold().dim());
+
                         if let Some(ref outpath) = out {
-                            let parent =
-                                outpath.parent().map(ToOwned::to_owned).ok_or_else(|| {
-                                    anyhow!("No valid parent directory for output file")
-                                })?;
+                            let name = name.map_or_else(|| hash.to_string(), |n| n);
+                            tokio::fs::create_dir_all(outpath).await?;
```
flub: probably want a `.context()` on this with a nice message

src/server.rs
```diff
@@ -31,15 +32,27 @@ const MAX_STREAMS: u64 = 10;
 
 pub type Database = Arc<HashMap<bao::Hash, Data>>;
 
+/// (Outboard, Data)
+// TODO: other option is to write the blobs data to a tempfile & store
+// in the main Database. Would need to add a "list of blobs" flag to `Data`
+// to make it clear that the data would deserialize to a `Collection` rather
+// than representing true blob data.
```
flub: The database is in memory only: why not store: `enum BlobOrCollection { Blob(Data), Collection(Collection) }`?  (ok, need better naming)

flub (commented):
Going to re-iterate my high-level comment from discord: I think this protocol is good for a TCP connection and a bit wasted on quic.  Whether we carry on with this to get the UX sorted first is still open of course, I'd be fine with this as first version of the protocol.

src/blobs.rs
ramfox: ahh yes thank you, I was being lazy about implementing `serde`

src/client.rs
ramfox: 😊

src/client.rs
ramfox: Oh yes, thank you! I was supposed to remove this test. 

Previously it was `if size != buffer.len()`, then it errored. But when transferring very small amounts of data, it is expected that the buffer `len` would be larger than `size`, so testing for equality isn't actually what we want.

src/client.rs
ramfox: this is just a "left over" comment that really translates to "add async decode to bao"

changing the comment

src/client.rs
ramfox: Okay, yes thank you I meant to bring this up.

I'm a bit confused about what we want here.

The docs for async `read_buf` say, when we read 0 bytes off the reader:

> This reader has reached its "end of file" and will likely no longer be able to produce bytes. Note that this does not mean that the reader will always no longer be able to produce bytes.

So is this saying that, for instance, if we have a not-so-fast connection, we may have 0 bytes to read off the stream, but may have more in the future? Or is it safe to assume that when we read 0 bytes, we are likely at EOF, and I'm being paranoid?

ramfox: This "likely" word is tripping me up about expectations.

src/server.rs
ramfox: lol too bad I basically used your naming

src/provider.rs
```diff
@@ -172,10 +254,16 @@ pub enum DataSource {
...
 }
 
-pub async fn create_db(data_sources: Vec<DataSource>) -> Result<Arc<HashMap<bao::Hash, Data>>> {
+// Creates a database of blobs (stored in outboard storage) and Collections, stored in memory.
+// Returns a the hash of the collection created by the given list of DataSources
+pub async fn create_db(data_sources: Vec<DataSource>) -> Result<(Database, bao::Hash)> {
     println!("Available Data:");
 
     let mut db = HashMap::new();
+    let mut blobs = Vec::new();
```
flub: You can call `Vec::with_capacity` (and for the HashMap above) using the length of `data_sources`

src/provider.rs
```diff
@@ -157,7 +194,52 @@ async fn handle_stream(
...
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
+enum SentStatus {
+    Sent,
+    NotFound,
+}
+
+async fn send_blob<W: AsyncWrite + Unpin>(
```
flub, line 231: this code should probably be merged with the blob branch in the loop of `handle_stream()` above.  a bit too repetitive.

src/blobs.rs
```diff
@@ -0,0 +1,45 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub(crate) struct Collection {
+    /// The name of this collection
+    pub(crate) name: String,
+    /// Links to the blobs in this collection
+    pub(crate) blobs: Vec<Blob>,
+    /// The total size of the raw_data referred to by all links
+    pub(crate) raw_size: usize,
```
flub: I tried not to nitpick on this, but I find it really hard to read.  I never know what that raw size applies to.  Can we name this `total_blobs_size` instead?  I don't mind being verbose.

src/blobs.rs
```diff
@@ -0,0 +1,45 @@
...
+#[serde(into = "SerdeBlob")]
+pub(crate) struct Blob {
+    /// The name of this blob of data
+    pub(crate) name: String,
+    /// The hash of the blob of data
+    pub(crate) hash: bao::Hash,
+}
+
+#[derive(Deserialize, Serialize, Clone)]
+struct SerdeBlob {
```
flub: I haven't tried this, but can this struct not be avoided by using `#[serde(serialize_with = "serialize_bao_hash")]` on the `Blob::hash` field?  You'd have to write `serialize_bao_hash` of course.

src/blobs.rs
```diff
@@ -0,0 +1,45 @@
...
+            hash: *b.hash.as_bytes(),
+        }
+    }
+}
+
+impl From<SerdeBlob> for Blob {
+    fn from(d: SerdeBlob) -> Self {
+        Self {
+            name: d.name,
+            hash: bao::Hash::from(d.hash),
```
dignifiedquire: can you file an issue with `bao` to add optional `serde` support?

src/protocol.rs
```diff
@@ -44,6 +44,16 @@ pub enum Res<'a> {
...
     },
+    /// Indicates that the given hash referred to a collection of multiple blobs
+    /// A stream of boa data that decodes to a `Collection` is sent as the next message,
+    /// followed by `Res::Found` responses, send in the order indicated in the `Collection`.
+    FoundCollection {
+        /// The size of the coming data in bytes, raw content size.
+        size: usize,
+        outboard: &'a [u8],
+        /// The size of the raw data we are planning to transfer
+        raw_transfer_size: usize,
```
flub: Just like in the `Collection` itself this name  keeps confusing me to no end, `total_blobs_size` would help me I think.

src/get.rs
```diff
@@ -119,82 +141,90 @@ pub fn run(hash: bao::Hash, opts: Options) -> impl Stream<Item = Result<Event>>
...
+                                        }
+                                     },
+                                     None => {
+                                        Err(anyhow!("server disconnected"))?;
+                                     }
+                                }
+                            }
+                        }
+                        // server is sending over a single blob
                         Res::Found { size, outboard } => {
```
dignifiedquire: We should get rid of this event, and always use the collection, to make the construction simpler

rklaehn:
Tried this out with the test directories from back when I implemented the HAMT. But nothing works because of some limits.

```
# a very large dir with tiny files
cargo run --release provide ../testdir_many_files/hamt1/*
- af1349b9f5f9a1a6a0404dea36dcc9499bcb25c9adc112b7cc9a93cae41f3262: 0 bytes
...
- af1349b9f5f9a1a6a0404dea36dcc9499bcb25c9adc112b7cc9a93cae41f3262: 0 bytes
Error: The serialize buffer is full

# a moderately large dir with tiny files
cargo run --release provide ../testdir_many_files/nohamt1/*
- af1349b9f5f9a1a6a0404dea36dcc9499bcb25c9adc112b7cc9a93cae41f3262: 0 bytes
...
- af1349b9f5f9a1a6a0404dea36dcc9499bcb25c9adc112b7cc9a93cae41f3262: 0 bytes
Error: The serialize buffer is full

# a dir with a few large files - fails when requesting:
Fetching: a4eae4b9cfd5d62076af83d07c0cf784133ebb2f3975fd2ea93ee8f04943e56c
[1/3] Connecting ...
[2/3] Requesting ...
Error: size too large: 5000000000 > 1073741824
```

Maybe we should increase the limits a bit...

rklaehn:
So it seems that around 1000 files is the current limit for the number of files.

I tried it out with 1000 empty files, and am getting this:

send side seems to work:
```
- af1349b9f5f9a1a6a0404dea36dcc9499bcb25c9adc112b7cc9a93cae41f3262: 0 bytes
...
- af1349b9f5f9a1a6a0404dea36dcc9499bcb25c9adc112b7cc9a93cae41f3262: 0 bytes

collection:
- 7ebc5acba68de920573dd4194ad661ce1858ff9ee546ebc91de9bc1924eea4ee
PeerID: e4e0e99e0efb496b0534a456eb6783153e5c7bd515aa44d196b3b210b15ef9da
```

but recv side:
```
❯ cargo run --release -- get --peer-id e4e0e99e0efb496b0534a456eb6783153e5c7bd515aa44d196b3b210b15ef9da 7ebc5acba68de920573dd4194ad661ce1858ff9ee546ebc91de9bc1924eea4ee > /dev/null
    Finished release [optimized] target(s) in 0.08s
     Running `target/release/sendme get --peer-id e4e0e99e0efb496b0534a456eb6783153e5c7bd515aa44d196b3b210b15ef9da 7ebc5acba68de920573dd4194ad661ce1858ff9ee546ebc91de9bc1924eea4ee`
Error: no more data available
```

Is it possible that empty files are a problem?

Update: tried this with an increasing number of empty files. 20 still works, ~40~ 30 causes the above problem.

src/blobs.rs
rklaehn: bao::Hash is just a reexport of blake3::Hash.

https://github.com/oconnor663/bao/blob/master/src/lib.rs#L46

So you would have to file this issue in blake3 directly. I did it here: https://github.com/BLAKE3-team/BLAKE3/issues/290

src/blobs.rs
rklaehn: There is also `#[serde(with = ...)` which allows you to define a module containing both serialize and deserialize. https://serde.rs/field-attrs.html#with

Worth a try, but the current way is also perfectly fine imho.

src/lib.rs
```diff
@@ -22,47 +23,158 @@ mod tests {
...
     async fn basics() -> Result<()> {
+        let filename = "hello_world";
+        let port: u16 = 4443;
+        single_file(filename, "hello world!".as_bytes(), false, port).await?;
+        single_file(filename, "hello world!".as_bytes(), true, port).await
+    }
+
+    // Run the test for a single file, with the option to transfer it as part of a collection or on
+    // its own
+    // TODO: random ports? rather than assigning?
```
dignifiedquire: yes: https://github.com/n0-computer/sendme/issues/49

src/lib.rs
```diff
@@ -22,47 +23,158 @@ mod tests {
...
+    async fn single_file(filename: &str, data: &[u8], wrapped: bool, port: u16) -> Result<()> {
```
dignifiedquire: In line with the above, I would like to see the `wrapped` option be removed, and have all files be always transferred as part of a collection

src/lib.rs
```diff
@@ -22,47 +23,158 @@ mod tests {
...
             peer_id: Some(peer_id),
         };
-        let stream = get::run(hash, opts);
+        let stream = get::run(collection_hash, opts);
+
+        tokio::pin!(stream);
+
+        // needs to iterate until `done`
         tokio::pin!(stream);
+        let mut i = 0;
```
dignifiedquire, line 151: is this used for anything?

src/provider.rs
```diff
@@ -172,10 +254,16 @@ pub enum DataSource {
     File(PathBuf),
 }
 
-pub async fn create_db(data_sources: Vec<DataSource>) -> Result<Arc<HashMap<bao::Hash, Data>>> {
+// Creates a database of blobs (stored in outboard storage) and Collections, stored in memory.
```
dignifiedquire: tripple slash for doc comments

src/provider.rs
```diff
@@ -187,20 +275,47 @@ pub async fn create_db(data_sources: Vec<DataSource>) -> Result<Arc<HashMap<bao:
...
-
-    Ok(Arc::new(db))
+    let c = Collection {
+        name: "collection".to_string(),
+        blobs,
+        raw_size,
+    };
+    blobs_encoded_size_estimate += c.name.len();
+
+    // NOTE: we can't use the postcard::MaxSize to estimate the encoding buffer size
```
dignifiedquire, line 334: wrote a comment regarding this issue: https://github.com/jamesmunns/postcard/issues/46#issuecomment-1403477151

dignifiedquire:
My main high level issue is that I want the individual file transfer to be removed, in favor of always sending a collection. The overhead should be minimal, and it makes the code a lot simpler.

dignifiedquire:
(sorry did not mean to close this)

rklaehn (changes requested):
See the comments I added when trying this out.

The low limits is an inconvenience, but the error I think is a genuine bug. See https://github.com/n0-computer/sendme/pull/48#issuecomment-1403414821

src/blobs.rs
flub: We should not send `usize` over to different machines.  Probably should be `u64`.

rklaehn: usize would also be a bit too small. It is plausible that you may want to send more than 4gb on a 32 bit machine.

src/get.rs
flub: The implication is that the hash for a blob can no longer verify the fetching of the collection.  So you need to fetch the collection (which should admittedly be small) without verifying.  However when you do have a collection you probably do want to verify it as it might be a huge thing.  So what tradeoff do we want to make here?  Fetch the collection without verifying as long as it is smaller than some size?

src/client.rs
ramfox: fixed by https://github.com/n0-computer/sendme/pull/51

src/blobs.rs
ramfox: Nitpick away! I feel naming for clarity is super important and am happy to have feedback on it.

src/blobs.rs
ramfox: ya, I was intimidated by doing `#[serde(with = ...)` because the `serde` docs for `Deserialize` describe having to also implement a `Vistitor`, which seemed needlessly complicated for translating a 32 byte array to a hash.  So I went for path of least resistance & used a `container-attr` rather than the `field-attr`.

But! I found a working example where someone described how you don't have to impl `Visitor` to get `Deserialize` to work: https://users.rust-lang.org/t/need-help-with-serde-deserialize-with/18374/5

So, using `#[serde(with = ...)` now!

src/blobs.rs
ramfox: What do I do when we have to do things like 
`buffer.split_to(usize)` and we have to "downcast" to `usize` anyway?

src/blobs.rs
ramfox: actually, the example without using a `Visitor` didn't work properly, just implemented it w/ a `Visitor` and works as expected now.

src/get.rs
ramfox: We currently use `Res::Found` to deliver the `outboard` encoded data & to ensure that we don't receive a `Re::NotFound` message instead. It's easy enough to error if we get a `Res::Found` for the first response, and leave `Res::Found` as a valid response when we are waiting for the blob data.

Which is how I have it set up in the latest commit.

If we really want to remove the `Res::Found` message, perhaps the protocol is changed to this:
- Receive `FoundCollection` or `NotFound`
- if `FoundCollection`, expect blobs in the order in which they are listed in the `Collection`
- receive length prefixed bao `outboard` encoded data
- receive length prefixed actual content
- verify content using outboard encoding & hash (that we have from the `Collection`)

My main question is what do we do in the case where the `Provider` does not have the content of a blob? Does it still send `Res::NotFound`, even though we will not be expecting a `Res` message on the `get` side? Or does it just close the connection, and the `getter` just reports an error, rather than a specific `NotFound` message?

src/blobs.rs
rklaehn: If you have to load the entire data into memory, your data can not be larger than `usize`. But you don't have to do that.

What you have to do if you e.g. want to handle a 1TB file on a 32bit / 4GB max architecture is to make sure that you never load more than 4GB in memory. But the overall file can still be larger than 4GB.

E.g. the std::io and tokio::io methods that deal with files are all working with u64. See for example https://doc.rust-lang.org/std/fs/fn.copy.html . It uses an internal buffer that is (much) less than 4GB, but the overall amount being copied can be much more than 4GB.

When writing code on 32 bit that deals with large files you have to constantly think about whether a certain 64 bit number is safe to downcast to usize or not.

src/get.rs
dignifiedquire: the protocol you described is what I was expecting, yes. 
I would abort if a blob is not found for now, given that the original collection found should only be sent if the provider has the whole collection

src/get.rs
```diff
@@ -139,82 +143,72 @@ pub fn run(hash: bao::Hash, token: AuthToken, opts: Options) -> impl Stream<Item
...
+                        // server is sending over a collection of blobs
+                        Res::FoundCollection { size, outboard, total_blobs_size } => {
+                            if total_blobs_size > MAX_DATA_SIZE {
+
+                                Err(anyhow!("size too large: {} > {}", total_blobs_size, MAX_DATA_SIZE))?;
                             }
+                            data_len = total_blobs_size;
 
-                            // TODO: avoid buffering
+                            yield Event::Requested { size: total_blobs_size };
```
dignifiedquire: should probably extend this with the notion that this is a collection not a blob

src/get.rs
```diff
@@ -139,82 +143,72 @@ pub fn run(hash: bao::Hash, token: AuthToken, opts: Options) -> impl Stream<Item
...
-                                reader.read_buf(&mut in_buffer).await?;
-                            }
+                            // read entire collection data into buffer :(
+                            let data = read_size_data(size, &mut reader, &mut in_buffer).await?;
 
-                            debug!("received data: {}bytes", in_buffer.len());
-                            if size != in_buffer.len() {
-                                Err(anyhow!("expected {} bytes, got {} bytes", size, in_buffer.len()))?;
+                            // decode the collection
+                            let collection = Collection::decode_from(data, outboard, hash).await?;
```
dignifiedquire, line 165: what about `yield`ing the details of the collection (e.g. it's name and number of blobs)? maybe add an event `Event::ReceivedCollection(Collection)`?

src/get.rs
```diff
@@ -223,3 +217,77 @@ pub fn run(hash: bao::Hash, token: AuthToken, opts: Options) -> impl Stream<Item
...
+// verified data to the `AsyncRead`er.
+async fn decode_data_to_reader(
+    data: Bytes,
+    outboard: &[u8],
+    hash: bao::Hash,
+) -> Result<(
+    Box<dyn AsyncRead + Unpin + Sync + Send + 'static>,
+    tokio::task::JoinHandle<Result<()>>,
+)> {
+    println!("decoding theses bytes: {:?}", data);
```
dignifiedquire: remove

src/get.rs
dignifiedquire: yeah after reviewing the latest state, I think you can leave the protocol as is, it's pretty clean now, thanks

rklaehn:
I still see a bug when trying to sync a directory containing 30 empty files:

provide side works:
```
❯ cargo run --release -- provide ../testdir_many_files/nohamt3/*
   Compiling zeroize v1.5.7
...
   
collection:
- cbf64610540529cf9f8ad4fd217fd6b8bf08421c6a230ff4c17c0b367b613b3c
PeerID: 321d91031dcd74227f16dd9c3574b146eedf44f24735eeb33723fb3036de35e1
Auth token: 49f398fabebe087e53f897e24928d131fd5b2f77e59fe521f112c0f3b5fb0e0f
```

but get side bails out:
```
sendme on  ramfox/blobs is 📦 v0.1.0 via 🦀 v1.66.0 
❯ cargo run --release -- get --peer-id 321d91031dcd74227f16dd9c3574b146eedf44f24735eeb33723fb3036de35e1 49f398fabebe087e53f897e24928d131fd5b2f77e59fe521f112c0f3b5fb0e0f cbf64610540529cf9f8ad4fd217fd6b8bf08421c6a230ff4c17c0b367b613b3c
    Finished release [optimized] target(s) in 0.11s
     Running `target/release/sendme get --peer-id 321d91031dcd74227f16dd9c3574b146eedf44f24735eeb33723fb3036de35e1 49f398fabebe087e53f897e24928d131fd5b2f77e59fe521f112c0f3b5fb0e0f cbf64610540529cf9f8ad4fd217fd6b8bf08421c6a230ff4c17c0b367b613b3c`
Fetching: cbf64610540529cf9f8ad4fd217fd6b8bf08421c6a230ff4c17c0b367b613b3c
[1/3] Connecting ...
[2/3] Requesting ...
Error: no more data available
```

src/blobs.rs
flub: To add to this, while right now we do horrible things like `buffer.split_to(entire_file_size)` we won't be doing this for too long I hope.

src/blobs.rs
```diff
@@ -0,0 +1,105 @@
...
+    where
+        D: Deserializer<'de>,
+    {
+        struct HashVisitor;
+
+        impl<'de> de::Visitor<'de> for HashVisitor {
+            type Value = bao::Hash;
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+                formatter.write_str("an array of bytes containing hash data")
```
flub:
```suggestion
                formatter.write_str("an array of 32 bytes containing hash data")
```

src/blobs.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// The name of this collection
+    pub(crate) name: String,
+    /// Links to the blobs in this collection
+    pub(crate) blobs: Vec<Blob>,
+    /// The total size of the raw_data referred to by all links
+    pub(crate) total_blobs_size: u64,
+}
+
+impl Collection {
+    pub async fn decode_from(data: Bytes, outboard: &[u8], hash: bao::Hash) -> Result<Self> {
```
flub, line 19: love this as an associated function :green_heart:

src/get.rs
```diff
@@ -223,3 +217,77 @@ pub fn run(hash: bao::Hash, token: AuthToken, opts: Options) -> impl Stream<Item
...
+                    let data = read_size_data(size, &mut reader, buffer).await?;
+                    decode_data_to_reader(data, outboard, hash).await
+                }
+            }
+        }
+        None => Err(anyhow!("server disconnected"))?,
+    }
+}
+
+// Returns an `AsyncRead` and a `JoinHandle`.
```
flub: `///`

src/blobs.rs
ramfox: Okay cool. For now, is the current set up reasonable, since we will most _not_ be reading the entire file to the buffer for much longer?

src/get.rs
ramfox: one day I'll remember this... :)

src/get.rs
ramfox: > I would abort if a blob is not found for now, given that the original collection found should only be sent if the provider has the whole collection

We currently do not check if each member of the collection exists at the provider before sending over the `FoundCollection` response. Should we refactor to do that, or is this case covered sufficiently in your mind now that we are keeping the `Res::Found` response?

dignifiedquire: at least a todo that we should check at the beginning

src/blobs.rs
ramfox: The conclusion made (in discord discussion) is to fix any issues in a follow-up, as rklaehn's upcoming planned work will likely eliminate many places where we currently have issues.

ramfox:
> I still see a bug when trying to sync a directory containing 30 empty files:
> 
> provide side works:
> 
> ```
> ❯ cargo run --release -- provide ../testdir_many_files/nohamt3/*
>    Compiling zeroize v1.5.7
> ...
>    
> collection:
> - cbf64610540529cf9f8ad4fd217fd6b8bf08421c6a230ff4c17c0b367b613b3c
> PeerID: 321d91031dcd74227f16dd9c3574b146eedf44f24735eeb33723fb3036de35e1
> Auth token: 49f398fabebe087e53f897e24928d131fd5b2f77e59fe521f112c0f3b5fb0e0f
> ```
> 
> but get side bails out:
> 
> ```
> sendme on  ramfox/blobs is 📦 v0.1.0 via 🦀 v1.66.0 
> ❯ cargo run --release -- get --peer-id 321d91031dcd74227f16dd9c3574b146eedf44f24735eeb33723fb3036de35e1 49f398fabebe087e53f897e24928d131fd5b2f77e59fe521f112c0f3b5fb0e0f cbf64610540529cf9f8ad4fd217fd6b8bf08421c6a230ff4c17c0b367b613b3c
>     Finished release [optimized] target(s) in 0.11s
>      Running `target/release/sendme get --peer-id 321d91031dcd74227f16dd9c3574b146eedf44f24735eeb33723fb3036de35e1 49f398fabebe087e53f897e24928d131fd5b2f77e59fe521f112c0f3b5fb0e0f cbf64610540529cf9f8ad4fd217fd6b8bf08421c6a230ff4c17c0b367b613b3c`
> Fetching: cbf64610540529cf9f8ad4fd217fd6b8bf08421c6a230ff4c17c0b367b613b3c
> [1/3] Connecting ...
> [2/3] Requesting ...
> Error: no more data available
> ```

This bug occurs because we don't have the "right" estimate for encoding overhead in postcard, when we `write_response` in `src/provider.rs`

However, it occurs to me that we probably shouldn't be sending the boa encoded data (the culprit for ballooning the postcard encoding overhead) in our response messages anyway, and should probably adjust the protocol regardless of how we better estimate the postcard overhead in our buffer.

src/blobs.rs
```diff
@@ -36,6 +37,18 @@ impl Collection {
             postcard::from_bytes(&data).context("failed to serialize Collection data")?;
         Ok(c)
     }
+
+    pub fn total_blobs_size(&self) -> u64 {
+        self.total_blobs_size
+    }
+
+    pub fn name(&self) -> String {
```
dignifiedquire: should return `&str`

dignifiedquire (approved):
one last nit, otherwise lgtm

PR #51, opened 2023-01-25
fix: Keep reading and respect EOF

flub:
We have two read loops that are a misbehaving a bit.  Make sure to
keep reading until we have the data, but bail on EOF.

### DIFF
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -89,7 +89,10 @@ pub async fn read_lp<'a, R: AsyncRead + futures::io::AsyncRead + Unpin, T: Deser
     let size = read_prefix(&mut reader, buffer).await?;
 
     while buffer.len() < size {
-        debug!("reading message {} {}", buffer.len(), size);
+        debug!("reading message, buffered {} of {size}", buffer.len());
+        if reader.read_buf(buffer).await? == 0 {
+            bail!("no more data available");
+        }
     }
     let response: T = postcard::from_bytes(&buffer[..size])?;
     debug!("read message of size {}", size);
@@ -106,7 +109,10 @@ pub async fn read_lp_data<R: AsyncRead + futures::io::AsyncRead + Unpin>(
     let size = read_prefix(&mut reader, buffer).await?;
 
     while buffer.len() < size {
-        reader.read_buf(buffer).await?;
+        debug!("reading data, buffered {} of {size}", buffer.len());
+        if reader.read_buf(buffer).await? == 0 {
+            bail!("no more data available");
+        }
     }
     let response = buffer.split_to(size);
     Ok(Some(response))


dignifiedquire (approved):

src/protocol.rs
```diff
@@ -89,7 +89,10 @@ pub async fn read_lp<'a, R: AsyncRead + futures::io::AsyncRead + Unpin, T: Deser
     let size = read_prefix(&mut reader, buffer).await?;
 
     while buffer.len() < size {
-        debug!("reading message {} {}", buffer.len(), size);
+        debug!("reading message, buffered {} of {size}", buffer.len());
+        if reader.read_buf(buffer).await? == 0 {
```
rklaehn, line 93: So 0 means "EOF for now", and we want to bail instead of hanging forever? :+1:

I guess `anyhow::ensure!(reader.read_buf(buffer).await? > 0);` would also work, but it has the disadvantage of putting the side effect in the check, so this is fine.

rklaehn (approved):

src/protocol.rs
flub: 0 means EOF afaik for all reasonable sources.  Though there's unfortunately some increadibly wiggly wording in the `AsyncReadExt` trait:

> [Return](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#return-1)
> 
> A nonzero n value indicates that the buffer buf has been filled in with n bytes of data > from this source. If n is 0, then it can indicate one of two scenarios:
> 
> 1. This reader has reached its “end of file” and will likely no longer be able to produce bytes. Note that this does not mean that the reader will always no longer be able to produce bytes.
> 2. The buffer specified had a remaining capacity of zero.

I'm honestly not sure how one is supposed to deal with the "Note that this does not mean that the reader will always no longer be able to produce bytes." bit.

flub: This seems to call for a `FusedAsyncRead`, but that doesn't exist.

flub: Though `AsyncRead` itself contains
> If no data was read (buf.filled().len() is unchanged), it implies that EOF has been reached.

Which is pretty clear.  So I have no idea why `AsyncReadExt::read_buf` is so cautious with the wording.

dignifiedquire: you can use other BufMut types in read_buf, some of which are not growable like BytesMut is, so a 0 read can also occur if the buffer is full

flub: Yes, that's their `2.`.  That weird wording in in `1.` though.

ramfox: thank you this was giving me nightmares: https://github.com/n0-computer/sendme/pull/48#discussion_r1086208885

flub: @ramfox oh, apologies I missed that comment.  I agree with you that this is just weird, and based on the `AsyncRead` trait itself I'd even say this "may produce more bytes in the future" bit is just wrong.

PR #52, opened 2023-01-25
feat: Create a Builder for the Provider

flub:
This is a nicer API to create a builder.  Crucially it can give all
the information to connect to a provider once it is started.
Something that wasn't quite possible before.

This is part of n0-computer/iroh#749 but is probably clearer split off first.

### DIFF
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -35,7 +35,7 @@ pub struct Builder {
 
 impl Builder {
     /// Creates a new builder for [`Provider`] using the given [`Database`].
-    pub fn new(db: Database) -> Self {
+    pub fn with_db(db: Database) -> Self {
         Self {
             bind_addr: "127.0.0.1:4433".parse().unwrap(),
             keypair: Keypair::generate(),
@@ -91,7 +91,6 @@ impl Builder {
             listen_addr,
             keypair: self.keypair,
             auth_token: self.auth_token,
-            // db: self.db,
             task,
         })
     }
@@ -129,7 +128,6 @@ pub struct Provider {
     listen_addr: SocketAddr,
     keypair: Keypair,
     auth_token: AuthToken,
-    // db: Database,
     task: JoinHandle<()>,
 }
 
@@ -138,7 +136,7 @@ impl Provider {
     ///
     /// Once the done with the builder call [`Builder::spawn`] to create the provider.
     pub fn builder(db: Database) -> Builder {
-        Builder::new(db)
+        Builder::with_db(db)
     }
 
     /// Returns the address on which the server is listening for connections.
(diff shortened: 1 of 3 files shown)


src/provider.rs
```diff
@@ -7,78 +7,99 @@ use bytes::{Bytes, BytesMut};
...
+        self
     }
 
-    pub async fn run(&mut self, opts: Options) -> Result<()> {
+    /// Spawns the [`Provider`] in a tokio task.
+    ///
+    /// This will create the underlying network server and spawn a tokio task accepting
+    /// connections.  The returned [`Provider`] can be used to control the task as well as
+    /// get information about it.
+    pub fn spawn(self) -> Result<Provider> {
```
dignifiedquire, line 79: I am personally not a big fan of this, I like if the caller has the ability to manage the spawned tasks.

dignifiedquire: but it is nice to have the `join` and `abort` being integrated..

dignifiedquire: Maybe keep the `run` method and only add `spawn`?

src/provider.rs
```diff
@@ -7,78 +7,99 @@ use bytes::{Bytes, BytesMut};
...
+    bind_addr: SocketAddr,
     keypair: Keypair,
     auth_token: AuthToken,
     db: Database,
 }
 
-impl Provider {
+impl Builder {
+    /// Creates a new builder for [`Provider`] using the given [`Database`].
     pub fn new(db: Database) -> Self {
```
dignifiedquire: lets call this `with_db` maybe?

src/provider.rs
```diff
@@ -7,78 +7,99 @@ use bytes::{Bytes, BytesMut};
...
-        debug!("\nlistening at: {:#?}", server.local_addr().unwrap());
+        let listen_addr = server.local_addr().unwrap();
+        let db2 = self.db.clone();
+        let task = tokio::spawn(async move { Self::run(server, db2, self.auth_token).await });
+
+        Ok(Provider {
+            listen_addr,
+            keypair: self.keypair,
+            auth_token: self.auth_token,
+            // db: self.db,
```
dignifiedquire: delete?

src/provider.rs
dignifiedquire: Oh I see this is on the builder :/ I'd rather have the ability to construct the provider and then call spawn/run on it, than directly on the builder tbh

flub: My attempt here is to have `Provider` be more RAII: I either have it and it is fully functional or it doesn't exist.  If we create `Provider` and need a separate call to `.run()` then you can have the `Provider` object in two different states.  That would result in e.g. `provider.listen_addr()` having to become `fn listen_addr(&self) -> Option<SockAddr>` instead of a plain `fn listen_addr(&self) -> SockAddr`.

So that was why I went with `Builder::spawn`.  And I'm still not sure I would like to give up on that.

dignifiedquire: lets try it

flub: PTAL: I'm going to stick with this version, I think the tradeoff of not getting to manage the task manually is worth the simpler layout.  But feel free to still disagree, I can live with the alternative as well.

dignifiedquire (approved):

PR #53, opened 2023-01-25
feat(provider): add persistent identity

### LINKED ISSUE #50: feat: provide server with a keypair
Currently every run generates a new keypair, which makes it a bit hard to use. Also most people running as a server would probably want a fixed peer-id unless in a one-off context.

### LINKED ISSUE #744: feat: provide server with a keypair
Currently every run generates a new keypair, which makes it a bit hard to use. Also most people running as a server would probably want a fixed peer-id unless in a one-off context.

dignifiedquire:
Closes n0-computer/iroh#744

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ ring = "0.16.20"
 rustls = { version = "0.20.8", default-features = false, features = ["dangerous_configuration"] }
 s2n-quic = { version = "1.14.0", default-features = false, features = [ "provider-tls-rustls", "provider-address-token-default" ] }
 serde = { version = "1.0.152", features = ["derive"] }
+ssh-key = { version = "0.5.1", features = ["ed25519", "std", "rand_core"] }
 tempfile = "3.3.0"
 thiserror = "1.0.38"
 tokio = { version = "1.24.1", features = ["full"] }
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -32,6 +32,9 @@ enum Commands {
         /// Auth token, defaults to random generated.
         #[clap(long)]
         auth_token: Option<String>,
+        /// If this path is provided and it exists, the private key ist read from this file and used, if it does not exist the private key will be persisted to this location.
```
Arqu, line 35: If this path is provided and it exists, the private key **ist** read from this file and used, if it does not exist the private key will be persisted to this location.

Ja man :D

Cargo.toml
```diff
@@ -27,6 +27,7 @@ ring = "0.16.20"
 rustls = { version = "0.20.8", default-features = false, features = ["dangerous_configuration"] }
 s2n-quic = { version = "1.14.0", default-features = false, features = [ "provider-tls-rustls", "provider-address-token-default" ] }
 serde = { version = "1.0.152", features = ["derive"] }
+ssh-key = { version = "0.5.1", features = ["ed25519", "std", "rand_core"] }
```
Arqu, line 30: Nice, for a moment thought this would depend on OpenSSL and it's bindings.

Arqu (approved):
Just a small nit. TYVM

src/main.rs
dignifiedquire: need some German in here ;) SendMich :P

PR #54, opened 2023-01-26
Change data encoding from outboard then data to bao slice encoding

rklaehn:
Quite curious about the perf of this brute force baseline version that just uses the SyncIoBridge.

Note that this will heavily conflict with https://github.com/n0-computer/sendme/pull/48 . So it will probably have to be rewritten completely.

### DIFF
--- a/src/get.rs
+++ b/src/get.rs
@@ -12,7 +12,9 @@ use tokio::io::{AsyncRead, AsyncReadExt};
 use tokio_util::io::SyncIoBridge;
 use tracing::debug;
 
-use crate::protocol::{read_lp_data, write_lp, AuthToken, Handshake, Request, Res, Response};
+use crate::protocol::{
+    ensure_buffer_size, read_lp_data, write_lp, AuthToken, Handshake, Request, Res, Response,
+};
 use crate::tls::{self, Keypair, PeerId};
 
 const MAX_DATA_SIZE: usize = 1024 * 1024 * 1024;
@@ -145,7 +147,11 @@ pub fn run(hash: bao::Hash, token: AuthToken, opts: Options) -> impl Stream<Item
                 Some(response_buffer) => {
                     let response: Response = postcard::from_bytes(&response_buffer)?;
                     match response.data {
-                        Res::Found { size } => {
+                        Res::Found => {
+                            // the bao slice encoding contains the overall data size as a le encoded u64
+                            // so we are guaranteed to have at least 8 bytes
+                            ensure_buffer_size(&mut reader, &mut in_buffer, 8).await?;
+                            let size = u64::from_le_bytes(in_buffer[..8].try_into().unwrap()) as usize;
                             yield Event::Requested { size };
 
                             // Need to read the message now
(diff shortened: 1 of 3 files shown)


src/get.rs
```diff
@@ -144,49 +145,33 @@ pub fn run(hash: bao::Hash, token: AuthToken, opts: Options) -> impl Stream<Item
                 Some(response_buffer) => {
                     let response: Response = postcard::from_bytes(&response_buffer)?;
                     match response.data {
-                        Res::Found { size, outboard } => {
+                        Res::Found { size } => {
                             yield Event::Requested { size };
 
                             // Need to read the message now
                             if size > MAX_DATA_SIZE {
                                 Err(anyhow!("size too large: {} > {}", size, MAX_DATA_SIZE))?;
```
rklaehn, line 159: Unrelated to this PR, but does this actually do anything? I mean, shouldn't it be `return Err` or something?

dignifiedquire: it works, see the `?` at the end. Not the best style though..

src/protocol.rs
```diff
@@ -36,28 +36,26 @@ pub struct Request {
...
     },
 }
 
-impl Res<'_> {
+impl Res {
     #[allow(clippy::len_without_is_empty)]
     pub fn len(&self) -> usize {
         match self {
-            Self::Found { outboard, .. } => outboard.len(),
+            Self::Found { size: _ } => 0,
```
dignifiedquire: this whole method can be avoided now, it always returns `0`

src/provider.rs
```diff
@@ -142,17 +143,32 @@ async fn handle_stream(
...
+                                file_reader,
+                                outboard_reader,
+                                0,
+                                size as u64,
+                            );
+                            let _copied = std::io::copy(&mut slice_extractor, &mut wrapper)?;
+                            std::io::Result::Ok(writer)
+                        })
+                        .await
+                        .unwrap()?;
```
dignifiedquire, line 166: should this be an `unwrap`?

src/get.rs
rklaehn: Ah, yes, saw that now.

src/provider.rs
rklaehn: nah, it shouldn't. Changed it.

src/protocol.rs
rklaehn: yes, removed it.

src/get.rs
```diff
@@ -151,8 +151,8 @@ pub fn run(hash: bao::Hash, token: AuthToken, opts: Options) -> impl Stream<Item
                             // the bao slice encoding contains the overall data size as a le encoded u64
                             // so we are guaranteed to have at least 8 bytes
                             ensure_buffer_size(&mut reader, &mut in_buffer, 8).await?;
-                            let size = u64::from_le_bytes(in_buffer[..8].try_into().unwrap()) as usize;
-                            yield Event::Requested { size };
+                            let size = u64::from_le_bytes(in_buffer[..8].try_into().unwrap());
+                            yield Event::Requested { size: usize::try_from(size)? };
```
dignifiedquire, line 155: the event should use  `u64` size as well

src/provider.rs
```diff
@@ -238,8 +238,8 @@ async fn write_response<W: AsyncWrite + Unpin>(
 ) -> Result<()> {
     let response = Response { id, data: res };
 
-    if buffer.len() < 20 + response.data.len() {
-        buffer.resize(20 + response.data.len(), 0u8);
+    if buffer.len() < 20 {
```
dignifiedquire, line 241: with this we can now use `MAX_SIZE` from postcard as there are no dynamic values anymore in the structs: https://docs.rs/postcard/latest/postcard/experimental/max_size/index.html

dignifiedquire: instead of the guessed `20`

rklaehn:
Replaced by #70

PR #55, opened 2023-01-26
refactor: use genawaiter instead of async-stream

dignifiedquire:
- genawaiter gives us more flexible generators, that can also be used to build `Stream`s as shown here
- syntax avoid nesting inside of macros

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,27 +89,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-stream"
-version = "0.3.3"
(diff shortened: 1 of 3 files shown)


rklaehn (approved):
I like it. I really don't like having pages of code go red in the IDE if you do something that is perfectly valid rust code but just not allowed in the subset of rust that is allowed in the try_stream macro...

image

flub (approved):
fewer macros! :tada:

rklaehn (approved):
Awesome. Much nicer than try_stream in terms of ergonomics while working on the code.

PR #57, opened 2023-01-26
refactor: compute outboard without reading the entire file

rklaehn:
this also makes possible handling >4gb on 32 bit systems, and makes sendme usable on machines with little memory.

Since we can now handle >4GB on 32 bit systems, also upgrade size to u64.

### DIFF
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -2,7 +2,7 @@ use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::{collections::HashMap, sync::Arc};
 
-use anyhow::{anyhow, bail, ensure, Result};
+use anyhow::{anyhow, bail, ensure, Context, Result};
 use bytes::{Bytes, BytesMut};
 use s2n_quic::stream::BidirectionalStream;
 use s2n_quic::Server as QuicServer;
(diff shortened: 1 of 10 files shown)


src/provider.rs
```diff
@@ -211,16 +211,46 @@ pub async fn create_db(data_sources: Vec<DataSource>) -> Result<Arc<HashMap<bao:
...
+
+                    // the length we have actually written, should be the same as the length of the file.
+                    let len2 = std::io::copy(&mut file, &mut encoder)?;
+                    // this can fail if the file was appended to during encoding.
+                    anyhow::ensure!(len == len2, "file changed during encoding");
+                    // this flips the outboard encoding from post-order to pre-order
+                    let hash = encoder.finalize()?;
+                    anyhow::Ok((hash, outboard))
+                })
+                .await??;
```
rklaehn: We could also parallelize these later.

src/provider.rs
```diff
@@ -211,16 +211,46 @@ pub async fn create_db(data_sources: Vec<DataSource>) -> Result<Arc<HashMap<bao:
...
                     path.display()
                 );
-                let data = tokio::fs::read(&path).await?;
-                let (outboard, hash) = bao::encode::outboard(&data);
-
-                println!("- {}: {}bytes", hash.to_hex(), data.len());
+                let path2 = path.clone();
+                // spawn a blocking task for computing the hash and outboard.
+                // pretty sure this is best to remain sync even once bao is async.
+                let (hash, outboard) = tokio::task::spawn_blocking(move || {
```
dignifiedquire: Maybe put this into a standalone function?

rklaehn: yes. makes sense.

src/provider.rs
rklaehn: I think parallelizing this does not make any sense whatsoever before the PR that allows you to send multiple files lands...

PR #58, opened 2023-01-26
fix: Do not write to a tempfile

### LINKED ISSUE #56: Using /dev/null as output file is broken
Now that sendme writes to a temprorary file during the transfer using `/dev/null` as output is broken.  That's not nice.

### LINKED ISSUE #743: Using /dev/null as output file is broken

flub:
Writing to a tempfile and then moving it breaks using e.g. /dev/null
as output and probably other such things.

This replaces it with a destructor which will clean up the partially
written file unless explicitly marked as completed.  Reliability-wise
this is as good as they tempfile solution as tempfile did also rely on
drop guards.

Closes n0-computer/iroh#743

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,6 +1,6 @@
 use std::{net::SocketAddr, path::PathBuf, str::FromStr};
 
-use anyhow::{anyhow, ensure, Context, Result};
+use anyhow::{ensure, Context, Result};
 use clap::{Parser, Subcommand};
 use console::style;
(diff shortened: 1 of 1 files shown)


Arqu (commented):
Lot's of criminal activity here. I'd love @dignifiedquire to also take a quick glance. LGTM otherwise.

dignifiedquire:
Closing in favor of n0-computer/sendme#59

PR #61, opened 2023-01-26
feat: Provide a single ticket to fetch a hash

flub:
This adds a ticket format which enables fetching a hash using a single "ticket" which contains all the connection details.  With an IPv4 address this just over 200 bytes long.

# Reviewing

- Probably want to start with `Ticket` in `provider.rs`
- The `protocol::serde_hash` module will need merging with #48 (thanks for the code :wink:)
- The semantics of the `get` command are a bit messy now, everything is optional and the code errors out manually.  Maybe we should have a `get-ticket` sub-command instead which would make the semantics a lot more obvious.
- Note this doesn't target the main branch but rather #52.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -40,15 +40,12 @@ enum Commands {
         key: Option<PathBuf>,
     },
     /// Fetch some data by hash.
-    ///
-    /// The hash must be provided by either the positional argument or can be provided by
-    /// the ticket.  When using a ticket the hash, peer ID, authentication token and provider address are all ignored and provided by the ticket.
     #[clap(about = "Fetch the data from the hash")]
     Get {
         /// The root hash to retrieve.
         hash: bao::Hash,
-        #[clap(long, short)]
         /// PeerId of the provider.
+        #[clap(long, short)]
         peer: PeerId,
         /// The authentication token to present to the server.
         #[clap(long)]
--- a/src/main.rs
+++ b/src/main.rs
@@ -290,10 +290,10 @@ async fn get_interactive(
                 mut reader,
                 name,
             } => {
-                println!("  {}", style(format!("Receiving {hash}...")).bold().dim());
+                let name = name.map_or_else(|| hash.to_string(), |n| n);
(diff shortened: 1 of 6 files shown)


dignifiedquire:
agreed on `GetTicket` or similar command

src/provider.rs
```diff
@@ -307,3 +323,69 @@ async fn write_response<W: AsyncWrite + Unpin>(
...
 }
+
+/// A token containing everything to get a file from the provider.
+///
+/// This token contains various things needed for getting a file from a provider:
+///
+/// - The *hash* to retrieve.
+/// - The *peer ID* identifying the provider.
+/// - The *socket address* the provider is listening on.
+/// - The *authentication token* with permission for the root hash.
```
dignifiedquire: I would expected these descriptions to be on the fields directly

src/main.rs
```diff
@@ -66,21 +75,33 @@ async fn main() -> Result<()> {
...
-            println!("Fetching: {}", hash.to_hex());
-            let mut opts = get::Options {
-                peer_id: Some(peer_id),
-                ..Default::default()
+            let (hash, token, opts) = if let Some(ticket) = ticket {
+                let opts = get::Options {
+                    addr: ticket.addr,
+                    peer_id: Some(ticket.peer),
+                };
+                (ticket.hash, ticket.token, opts)
```
dignifiedquire: Maybe instead of taking these apart, always pass a ticket to the api, and create one from parts in the case they are passed in individually? (not sure this is better, just a thought)

src/provider.rs
flub: Yes, makes sense.  Originally this was opaque so I didn't.  But it made more sense with the fields as pub in the end.  Will change it tomorrow.

src/main.rs
flub: I did consider that, but I don't think I want the ticket to go far into the API, at least for now.  So I settled on this being not too ugly.

dignifiedquire (approved):

PR #62, opened 2023-01-26
refactor: avoid boxing and simplify buffer reads

dignifiedquire:


dignifiedquire:
According to @Arqu and his tests, this fixes n0-computer/iroh#742 :)

src/get.rs
```diff
@@ -140,70 +139,30 @@ pub fn run(hash: bao::Hash, token: AuthToken, opts: Options) -> impl Stream<Item
         // 3. Read response
         {
             debug!("reading response");
-            let mut in_buffer = BytesMut::with_capacity(1024);
+            let mut in_buffer = BytesMut::with_capacity(Response::POSTCARD_MAX_SIZE);
```
Arqu, line 142: How big is this. Might be a memory hog on high RPS.

dignifiedquire: this is a fixed small number, just the maximum number of bytes in the response encoding, which has no variable sized fields anymore

PR #68, opened 2023-01-27
feature: incrementally compute outboard

rklaehn:
...instead of loading the whole thing into memory. This means that we can work with files that are larger than available memory.

Replaces #57

### DIFF
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -2,7 +2,7 @@ use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::{collections::HashMap, sync::Arc};
 
-use anyhow::{anyhow, bail, ensure, Result};
+use anyhow::{anyhow, bail, ensure, Context, Result};
 use bytes::{Bytes, BytesMut};
 use s2n_quic::stream::BidirectionalStream;
 use s2n_quic::Server as QuicServer;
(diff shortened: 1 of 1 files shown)


src/provider.rs
```diff
@@ -282,6 +282,42 @@ pub enum DataSource {
...
+    let mut outboard = Vec::with_capacity(outboard_size);
+
+    // copy the file into the encoder. Data will be skipped by the encoder in outboard mode.
+    let outboard_cursor = std::io::Cursor::new(&mut outboard);
+    let mut encoder = bao::encode::Encoder::new_outboard(outboard_cursor);
+
+    // the length we have actually written, should be the same as the length of the file.
+    let len2 = std::io::copy(&mut file, &mut encoder)?;
+    // this can fail if the file was appended to during encoding.
+    anyhow::ensure!(len == len2, "file changed during encoding");
```
dignifiedquire: We import these macros in the other cases

src/provider.rs
```diff
@@ -301,19 +337,24 @@ pub async fn create_db(data_sources: Vec<DataSource>) -> Result<(Database, bao::
...
-                let (outboard, hash) = bao::encode::outboard(&data);
-
-                println!("- {}: {} bytes", hash.to_hex(), data.len());
+                // spawn a blocking task for computing the hash and outboard.
+                // pretty sure this is best to remain sync even once bao is async.
+                let path2 = path.clone();
+                let (hash, outboard) =
+                    tokio::task::spawn_blocking(move || compute_outboard(path2)).await??;
+
+                anyhow::ensure!(outboard.len() >= 8);
```
dignifiedquire: same here

dignifiedquire: Also please add an error message

dignifiedquire: I wonder if this should just be a `debug_assert`

rklaehn: I like keeping the anyhow:: prefix, but will adapt it to be consistent...

src/provider.rs
```diff
@@ -282,6 +282,42 @@ pub enum DataSource {
...
+///
+/// It is assumed that the file is not modified while this is running.
+///
+/// If it is modified while or after this is running, the outboard will be
+/// invalid, so any attempt to compute a slice from it will fail.
+///
+/// If the size of the file is changed while this is running, an error will be
+/// returned.
+fn compute_outboard(path: PathBuf) -> anyhow::Result<(blake3::Hash, Vec<u8>)> {
+    let mut file = std::fs::File::open(path)?;
```
dignifiedquire: should probably wrap the file into a `BufReader` to improve read speed

rklaehn: ✅

dignifiedquire (approved):

PR #70, opened 2023-01-27
Change data encoding from outboard then data to bao slice encoding

rklaehn:
...and use the async slice decoder.

Replaces #54

Note that this reintroduces forwarding the data via a duplex stream, which is probably not great for perf. This is needed because we are reading the blobs one by one from one stream and therefore need to thread the reader through the reading function.

Not sure how to best get rid of it. It seems to be inherent in the current protocol that you have to read one by one. Maybe pass in the target AsyncWrite?


src/bao_slice_decoder.rs
```diff
@@ -0,0 +1 @@
+
```
dignifiedquire, line 8: that's a sad file

rklaehn: the best code is no code...

dignifiedquire: taken to the MAX

src/blobs.rs
```diff
@@ -16,23 +16,13 @@ pub struct Collection {
...
-            // TODO: write & use an `async decoder`
-            let read = decoder
-                .read(&mut buf)
-                .context("hash of Collection data does not match")?;
-            if read == 0 {
-                break;
-            }
-        }
+        let mut decoder = bao::decode::Decoder::new(std::io::Cursor::new(&encoded[..]), &hash);
+        let mut data = Vec::new();
```
flub: Is there a size hint we can use here for `Vec::with_capacity`?

src/blobs.rs
```diff
@@ -16,23 +16,13 @@ pub struct Collection {
 }
 
 impl Collection {
-    pub async fn decode_from(data: Bytes, outboard: &[u8], hash: bao::Hash) -> Result<Self> {
-        // TODO: avoid copy
-        let outboard = outboard.to_vec();
+    pub async fn decode_from(encoded: Bytes, hash: bao::Hash) -> Result<Self> {
```
flub: There's no `.await` in this function body?  That seems suspicious.

src/get.rs
```diff
@@ -179,28 +179,29 @@ async fn inner_producer(
...
 
                         co.yield_(Ok(Event::ReceivedCollection(collection.clone())))
                             .await;
 
                         // expect to get blob data in the order they appear in the collection
                         for blob in collection.blobs {
-                            let (blob_reader, task) =
-                                handle_blob_response(blob.hash, &mut reader, &mut in_buffer)
-                                    .await?;
+                            let (blob_reader, reader1) =
```
flub: nit: it would help me if `reader1` still had `task` in it's name, `e.g. reader_task` or so.

src/provider.rs
```diff
@@ -227,6 +229,15 @@ async fn handle_stream(db: Database, token: AuthToken, stream: BidirectionalStre
                     Some(BlobOrCollection::Collection((outboard, data))) => {
                         debug!("found collection {}", name.to_hex());
 
+                        let mut extractor = SliceExtractor::new_outboard(
+                            std::io::Cursor::new(&data[..]),
+                            std::io::Cursor::new(&outboard[..]),
+                            0,
+                            data.len() as u64,
+                        );
+                        let mut encoded = Vec::new();
```
flub: Any size hint we can use?

flub (approved):

src/blobs.rs
rklaehn: Yes, we can use encoded_size since we know the size.

src/get.rs
rklaehn: obsoleted by digs pr.

src/blobs.rs
rklaehn: It is using the sync encoder, since the stuff is in mem anyway. Made it a sync fn. We can make async again once we find a reason to make it async.

dignifiedquire (approved):

PR #71, opened 2023-01-27
feat: remove varints

dignifiedquire:
and use `take` to have cleaner protocol reads

### DIFF
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -100,12 +100,13 @@ pub async fn read_lp<'a, R: AsyncRead + Unpin, T: Deserialize<'a>>(
 
     let size = usize::try_from(size)?;
     let mut read = 0;
-    while buffer.len() < size {
+    while read != size {
         let r = reader.read_buf(buffer).await?;
-        ensure!(r != 0, "no more data available");
         read += r;
+        if r == 0 {
+            break;
+        }
     }
-    ensure!(size == read, "expected to read {} but read {}", size, read);
     let response: T = postcard::from_bytes(&buffer[..size])?;
     debug!("read message of size {}", size);
 
@@ -123,12 +124,13 @@ pub async fn read_size_data<R: AsyncRead + Unpin>(
     let mut reader = reader.take(size);
     let size = usize::try_from(size)?;
     let mut read = 0;
-    while buffer.len() < size {
+    while read != size {
         let r = reader.read_buf(buffer).await?;
-        ensure!(r != 0, "no more data available");
         read += r;
+        if r == 0 {
+            break;
+        }
     }
-    ensure!(size == read, "expected to read {} but read {}", size, read);
     debug!("finished reading");
     Ok(buffer.split_to(size).freeze())
 }
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -1,7 +1,7 @@
 use std::fmt::Display;
 use std::str::FromStr;
 
-use anyhow::{bail, ensure, Result};
+use anyhow::{ensure, Result};
 use bytes::{Bytes, BytesMut};
 use postcard::experimental::max_size::MaxSize;
 use serde::{Deserialize, Serialize};
@@ -81,9 +81,8 @@ pub async fn write_lp<W: AsyncWrite + Unpin>(writer: &mut W, data: &[u8]) -> Res
     );
 
     // send length prefix
-    let mut buffer = [0u8; 10];
-    let lp = unsigned_varint::encode::u64(data.len() as u64, &mut buffer);
-    writer.write_all(lp).await?;
+    let data_len = data.len() as u64;
+    writer.write_u64_le(data_len).await?;
 
     // write message
     writer.write_all(data).await?;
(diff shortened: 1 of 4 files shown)


src/protocol.rs
```diff
@@ -81,29 +81,31 @@ pub async fn write_lp<W: AsyncWrite + Unpin>(writer: &mut W, data: &[u8]) -> Res
     );
 
     // send length prefix
-    let mut buffer = [0u8; 10];
-    let lp = unsigned_varint::encode::u64(data.len() as u64, &mut buffer);
-    writer.write_all(lp).await?;
+    let data_len = data.len() as u64;
+    writer.write_u64_le(data_len).await?;
```
rklaehn, line 85: not that it matters that much, but why le?

src/protocol.rs
```diff
@@ -112,53 +114,41 @@ pub async fn read_lp<'a, R: AsyncRead + futures::io::AsyncRead + Unpin, T: Deser
...
+    debug!("finished reading");
+    Ok(buffer.split_to(size).freeze())
 }
 
 /// Return a buffer of the data, based on the length prefix, from the given source.
 /// The new buffer is split off from the buffer that is passed in the function.
-pub async fn read_lp_data<R: AsyncRead + futures::io::AsyncRead + Unpin>(
+pub async fn read_lp_data<R: AsyncRead + Unpin>(
     mut reader: R,
     buffer: &mut BytesMut,
```
rklaehn, line 142: So do we still need this annoying buffer at all?

The issue I have is that when handing the remaining reader to the AsyncSliceDecoder, I have to use chain to take these bytes into account.

But when I do that I have no way to get the reader out again, meaning that I can only read one item...

rklaehn (approved):

src/protocol.rs
flub, line 85: Little-endian does offend the greybeard hiding in me a little.  Network byte-order has always been big-endian.  Though it is true that most mainline systems are little-endian these days.

dignifiedquire: mostly a personal preference

src/protocol.rs
dignifiedquire: There should be no additional bytes in here anymore, see the `.take()` it now only reads the acutal required bytes, not more

src/protocol.rs
rklaehn: bao also uses le, so I guess it is consistent. Also I don't really mind.

src/protocol.rs
```diff
@@ -81,29 +81,31 @@ pub async fn write_lp<W: AsyncWrite + Unpin>(writer: &mut W, data: &[u8]) -> Res
...
+    let size = usize::try_from(size)?;
+    let mut read = 0;
     while buffer.len() < size {
-        debug!("reading message, buffered {} of {size}", buffer.len());
-        if reader.read_buf(buffer).await? == 0 {
-            bail!("no more data available");
-        }
+        let r = reader.read_buf(buffer).await?;
+        ensure!(r != 0, "no more data available");
+        read += r;
```
flub: Shouldn't one read to EOF if you're using `.take()`?

src/protocol.rs
```diff
@@ -81,29 +81,31 @@ pub async fn write_lp<W: AsyncWrite + Unpin>(writer: &mut W, data: &[u8]) -> Res
...
     }
+    ensure!(size == read, "expected to read {} but read {}", size, read);
```
flub: I wouldn't leave this in for prod code.  It really should be held up by the code just above already.

src/provider.rs
```diff
@@ -312,7 +311,8 @@ async fn send_blob<W: AsyncWrite + Unpin>(
             debug!("writing data");
             let file = tokio::fs::File::open(&path).await?;
             let mut reader = tokio::io::BufReader::new(file);
-            tokio::io::copy(&mut reader, &mut writer).await?;
+            tokio::io::copy_buf(&mut reader, &mut writer).await?;
```
flub, line 314: lol, love the lack of docs on this...

src/protocol.rs
dignifiedquire: ups, yes

src/protocol.rs
flub: heh, i had the same comment.  In theory these days with most machines being LE there is less byte-swapping involved, which can be an argument for it (and I assumed it was, turns out it wasn't even that :smile: )

src/protocol.rs
dignifiedquire: fixed

src/protocol.rs
dignifiedquire: fixed

src/provider.rs
dignifiedquire: yeah, so by reading the code, the TLDR is, use it if your sources provide AsyncBufRead, as that allows you to save on the additional buffer that `copy` creates.

src/protocol.rs
dignifiedquire: 🤣

flub (approved):

PR #78, opened 2023-01-30
refactor: switch to callbacks instead of events

dignifiedquire:
This allows avoiding the double copy when receiving data.

### DIFF
--- a/src/get.rs
+++ b/src/get.rs
@@ -200,13 +200,9 @@ where
 
 /// Read next response, and if `Res::Found`, reads the next blob of data off the reader.
 ///
-/// Returns an `AsyncReader` and a `JoinHandle`.
-///
+/// Returns an `AsyncReader`
 /// The `AsyncReader` can be used to read the content.
-///
-/// The `JoinHandle` task must be `await`-ed to get back the reader to read the next message.
 async fn handle_blob_response<
-    'a,
     R: AsyncRead + futures::io::AsyncRead + Send + Sync + Unpin + 'static,
 >(
     hash: bao::Hash,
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -243,31 +256,47 @@ async fn get_interactive(
...
 
     let pb = ProgressBar::hidden();
-    let stream = get::run(hash, token, opts);
-    tokio::pin!(stream);
-    while let Some(event) = stream.next().await {
-        trace!("client event: {:?}", event);
-        match event? {
-            get::Event::Connected => {
-                out_writer.println(format!("{} Requesting ...", style("[2/3]").bold().dim()));
+    let stats = get::run(
```
rklaehn, line 269: I think readability would be better if the three callbacks were not defined inline

rklaehn (approved):

src/get.rs
```diff
@@ -186,22 +147,17 @@ async fn inner_producer(
...
-                                reader: blob_reader,
-                                name: Some(blob.name),
-                            }))
-                            .await;
+                            let blob_reader =
+                                on_blob(blob.hash, blob_reader, Some(blob.name)).await?;
+                            reader = blob_reader.into_inner();
                             // await the completion of the copying. Only then can we get back the reader.
-                            reader = reader1.await??;
+                            // reader = reader1.await??;
```
flub, line 160: remove?

src/get.rs
```diff
@@ -65,68 +65,29 @@ pub struct Stats {
...
+    on_connected: A,
+    mut on_collection: B,
+    mut on_blob: C,
+) -> Result<Stats>
+where
+    A: FnOnce() -> FutA,
+    FutA: Future<Output = Result<()>>,
+    B: FnMut(Collection) -> FutB,
+    FutB: Future<Output = Result<()>>,
+    C: FnMut(bao::Hash, AsyncSliceDecoder<ReceiveStream>, Option<String>) -> FutC,
```
flub, line 81: Maybe also pass in the name of the blob if known as part of a collection?  Otherwise I can see the consumers having to map this separately anyway.

I guess same could be said about collection name.

flub (approved):
I like the API much less, but if the duplex really is that much slower I guess it's a fair tradeoff.

I also wonder if an alternative would be to have an API which has to return a writer instead of a future, something like `fn on_blob(hash: boa::Hash) -> impl AsyncWriter`.  Would that be worse or similar performance-wise?

flub:
> I also wonder if an alternative would be to have an API which has to return a writer instead of a future, something like `fn on_blob(hash: boa::Hash) -> impl AsyncWriter`. Would that be worse or similar performance-wise?

I guess that would still have to be an `async fn` though.

PR #81, opened 2023-01-31
fix(bin): ensure progressbar works for stdout

dignifiedquire:
also redirects logs to STDERR

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -112,7 +112,7 @@ const PROGRESS_STYLE: &str =
 #[tokio::main(flavor = "multi_thread")]
 async fn main() -> Result<()> {
     tracing_subscriber::registry()
-        .with(fmt::layer())
+        .with(fmt::layer().with_writer(std::io::stderr))
         .with(EnvFilter::from_default_env())
         .init();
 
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -266,6 +266,19 @@ async fn get_interactive(
         .await;
 
     let pb = ProgressBar::hidden();
+    pb.enable_steady_tick(std::time::Duration::from_millis(1));
```
rklaehn: Isn't 1 millisecond a bit much? I would think that 50ms or even 500ms would still be fine.

rklaehn (approved):

src/main.rs
dignifiedquire: you are right, set to 50ms for now 🤷

src/main.rs
```diff
@@ -333,18 +342,18 @@ async fn get_interactive(
                 })
                 .await??;
                 let file = tokio::fs::File::from_std(dup);
-                let out = tokio::io::BufWriter::new(file);
-                // wrap for progress bar
-                let mut wrapped_out = pb.wrap_async_write(out);
-                tokio::io::copy(&mut reader, &mut wrapped_out).await?;
+                let mut file_buf = tokio::io::BufWriter::new(file);
```
flub, line 347: The default buffer size is only 8KiB.  I think on most modern systems you can get a lot faster when writing to disk with something larger.  IIRC gnu coreutils defaults to 128KiB which they deem a good compromise for most cases, so I think that's a more reasonable default.

flub (approved):

src/main.rs
dignifiedquire: We have a bunch of these lying around, so lets change that/test that independently

PR #87, opened 2023-01-31
feat(provider): add events

### LINKED ISSUE #80: feat: Provider to know when a transfer finishes
Previously in the deltachat iroh-share usecase the provider would know when a transfer completed and would then stop the server-side.  I think sendme also needs an API for the provider to tell when a transfer is finished.

One way would be an event like there used to be on the client side that says `enum ProviderEvent { ClientConnected, HashRequested, TransferAborted, TransferCompleted}` could do.  Though strictly I only need the last two of that enum.

### LINKED ISSUE #734: feat: Provider to know when a transfer finishes
Previously in the deltachat iroh-share usecase the provider would know when a transfer completed and would then stop the server-side.  I think sendme also needs an API for the provider to tell when a transfer is finished.

One way would be an event like there used to be on the client side that says `enum ProviderEvent { ClientConnected, HashRequested, TransferAborted, TransferCompleted}` could do.  Though strictly I only need the last two of that enum.

dignifiedquire:
Closes n0-computer/iroh#734

### DIFF
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -12,6 +12,7 @@ use s2n_quic::stream::BidirectionalStream;
 use s2n_quic::Server as QuicServer;
 use serde::{Deserialize, Serialize};
 use tokio::io::{AsyncWrite, AsyncWriteExt};
+use tokio::sync::broadcast;
 use tokio::task::{JoinError, JoinHandle};
 use tokio_util::io::SyncIoBridge;
 use tracing::{debug, warn};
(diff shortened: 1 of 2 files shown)


src/provider.rs
```diff
@@ -143,6 +162,31 @@ pub struct Provider {
     keypair: Keypair,
     auth_token: AuthToken,
     task: JoinHandle<()>,
+    events: broadcast::Sender<Event>,
+    // Dummy receiver to ensure we don't error.
+    _events_receiver: broadcast::Receiver<Event>,
```
flub: I think I'd prefer not to store this, there's little value, on send we can call `events.send().ok()` and all is fine as we don't mind if anyone is observing it or not.

There's even an optimisation in the sender than does not bother trying to store the message if there are no receivers.  So that's another (tiny) reason not to store this receiver.

src/provider.rs
```diff
@@ -143,6 +162,31 @@ pub struct Provider {
...
+}
+
+/// Events emitted by the [`Provider`] informing about the current status.
+#[derive(Debug, Clone)]
+pub enum Event {
+    Running,
```
flub: Unless an event receiver is integrated into the `Builder` you'll never be able to observe this event I reckon.  Obviously you can now blame me for not having control over the `.run()` call as a user as that would have avoided this :smile:.  So going back to that design is still an option I guess.

Also, if there's a `Running` event I'd also expect a `Stopped` event or similar.  Honestly I think the easiest is to remove this event and call it a day, not sure when this would be very important to have.

flub (approved):

src/provider.rs
dignifiedquire: fixed

src/provider.rs
dignifiedquire: fixed

flub (approved):

PR #89, opened 2023-02-01
Remove double sizes - second attempt

rklaehn:
I was overcomplicating this. I thought you need the size for the slice decoder, but you don't. If you set max to u64::MAX, it will just read to the end and not more.

The way it works is that it reads the 8 byte header first, and then truncates the range to the valid range.

So all that is needed to remove the sizes is to... just remove the sizes.

### DIFF
--- a/src/blobs.rs
+++ b/src/blobs.rs
@@ -1,7 +1,4 @@
-use std::io::Read;
-
 use anyhow::{Context, Result};
-use bytes::Bytes;
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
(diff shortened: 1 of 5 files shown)


src/blobs.rs
```diff
@@ -16,16 +13,9 @@ pub struct Collection {
...
 impl Collection {
-    pub fn decode_from(encoded: Bytes, hash: bao::Hash) -> Result<Self> {
-        // verify that the content of data matches the expected hash
-        let mut decoder = bao::decode::Decoder::new(std::io::Cursor::new(&encoded[..]), &hash);
-        // decoded size can be at most encoded size
-        let mut data = Vec::with_capacity(encoded.len());
-        decoder
-            .read_to_end(&mut data)
-            .context("hash of Collection data does not match")?;
+    pub fn deserialize_from(data: &[u8]) -> Result<Self> {
```
flub: nit: `from_bytes`?  I honestly don't know, maybe that would be more consistent, maybe not.

src/get.rs
```diff
@@ -218,9 +213,9 @@ async fn handle_blob_response<
                 // blob data not found
                 Res::NotFound => Err(anyhow!("data for {} not found", hash.to_hex()))?,
                 // next blob in collection will be sent over
-                Res::Found { size } => {
+                Res::Found => {
                     assert!(buffer.is_empty());
-                    let decoder = AsyncSliceDecoder::new(reader, hash, 0, size);
+                    let decoder = AsyncSliceDecoder::new(reader, hash, 0, u64::MAX);
```
flub, line 218: Is it worth adding a comment saying why `u64::MAX` is fine?  Or write a nice doc comment on `AsyncSliceDecoder::new` that including why you can pass `u64::MAX`.

flub (approved):

src/blobs.rs
rklaehn: affirmative!

src/get.rs
rklaehn: :+1: . This is on the way out anyway, but I will add it in bao as well.

flub (approved):

PR #90, opened 2023-02-01
perf: wrap the SliceExtractor in a BufReader when encoding

rklaehn:
Buffer size is chosen to be 2^16, so about what I think an ideal bao block size would be. Might be better to make this even bigger, but the default size of BufReader is definitely too small.

This is WIP since it still needs an easy way to see if it makes a difference. Thinking about having a large test that is disabled by default and that you can run manually, or even flamegraph it.

### DIFF
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -325,13 +325,14 @@ async fn send_blob<W: AsyncWrite + Unpin + Send + 'static>(
                 let file_reader = std::fs::File::open(&path)?;
                 let outboard_reader = std::io::Cursor::new(outboard);
                 let mut wrapper = SyncIoBridge::new(&mut writer);
-                let mut slice_extractor = bao::encode::SliceExtractor::new_outboard(
+                let slice_extractor = bao::encode::SliceExtractor::new_outboard(
                     file_reader,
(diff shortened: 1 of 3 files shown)


src/provider.rs
```diff
@@ -325,13 +325,14 @@ async fn send_blob<W: AsyncWrite + Unpin + Send + 'static>(
...
                 let mut wrapper = SyncIoBridge::new(&mut writer);
-                let mut slice_extractor = bao::encode::SliceExtractor::new_outboard(
+                let slice_extractor = bao::encode::SliceExtractor::new_outboard(
                     file_reader,
                     outboard_reader,
                     0,
                     size,
                 );
-                let _copied = std::io::copy(&mut slice_extractor, &mut wrapper)?;
+                let mut reader = std::io::BufReader::with_capacity(1 << 16, slice_extractor);
```
dignifiedquire, line 333: please make these constants with documentation as to why they are this size.

rklaehn: Yes, would do if this would make a difference. But it turns out that it does not.

rklaehn:
Did not make a difference in measurements...

PR #93, opened 2023-02-01
ref: Take the collection hash from the db creation

flub:
The database now returns the hash of the collection created.  There's
no need to go manually looking for this anymore, just use it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -8,7 +8,7 @@ use indicatif::{
 };
 use is_terminal::IsTerminal;
 use sendme::protocol::AuthToken;
-use sendme::provider::{BlobOrCollection, Ticket};
+use sendme::provider::Ticket;
 use tokio::io::AsyncWriteExt;
 use tokio::sync::Mutex;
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -191,15 +191,7 @@ async fn main() -> Result<()> {
...
-            let (db, _) = provider::create_db(sources).await?;
-            let hash = db
-                .iter()
-                .filter_map(|(hash, item)| match item {
-                    BlobOrCollection::Collection(_) => Some(hash),
-                    _ => None,
-                })
-                .next()
-                .copied();
+            let (db, hash) = provider::create_db(sources).await?;
```
dignifiedquire: maybe we should rename this to sth like `create_collection` or sth like this

flub: done

flub: I think the API to build this will continue to evolve a bit anyway.  See also https://github.com/n0-computer/sendme/issues/92

ramfox (approved):

PR #97, opened 2023-02-02
feat: pass in custom names for blobs

### LINKED ISSUE #92: Allow setting blob name when creating provider database
Currently creating the provider database is done by supplying a number of file paths.  The protocol has the concept of a blob name and this is taken to always be the filename part of the full path.

Some usecases, e.g. deltachat, require more control over the blob names so that they can introduce namespaces outside of the control of the filesystem that the receiver can utilise.

### LINKED ISSUE #727: Allow setting blob name when creating provider database
Currently creating the provider database is done by supplying a number of file paths.  The protocol has the concept of a blob name and this is taken to always be the filename part of the full path.

Some usecases, e.g. deltachat, require more control over the blob names so that they can introduce namespaces outside of the control of the filesystem that the receiver can utilise.

ramfox:
closes n0-computer/iroh#727 

I'm not too sure about the API for this, but it was the path of least resistance to see if this even solves for the use case

### DIFF
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -425,10 +425,30 @@ pub enum DataSource {
     /// A blob of data originating from the filesystem. The name of the blob is derived from
     /// the filename.
     File(PathBuf),
-    /// NamedFile is treated the same as [`DataSource::File`], except you can pass in a custom name
-    /// for the blob. If `name` is empty, it will derive the blob name from the `path`, in the same
-    /// way that we derive the blob name for [`DataSource::File`].
+    /// NamedFile is treated the same as [`DataSource::File`], except you can pass in a custom
+    /// name. Passing in the empty string will explicitly _not_ persist the filename.
     NamedFile { path: PathBuf, name: String },
+}
+
+impl DataSource {
+    pub fn new(path: PathBuf) -> Self {
+        DataSource::File(path)
+    }
+    pub fn with_name(path: PathBuf, name: String) -> Self {
+        DataSource::NamedFile { path, name }
+    }
+}
+
+impl From<PathBuf> for DataSource {
+    fn from(value: PathBuf) -> Self {
+        DataSource::new(value)
+    }
+}
+
+impl From<&std::path::Path> for DataSource {
+    fn from(value: &std::path::Path) -> Self {
+        DataSource::new(value.to_path_buf())
+    }
 }
 
 /// Synchronously compute the outboard of a file, and return hash and outboard.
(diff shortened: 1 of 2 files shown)


src/provider.rs
```diff
@@ -471,39 +477,43 @@ pub async fn create_collection(data_sources: Vec<DataSource>) -> Result<(Databas
...
+                size,
+            }),
+        );
+        total_blobs_size += size;
+        // if the given name is the empty `String` or `None`, use the filename from the
+        // given path as the name
+        let name = name.filter(|n| !n.is_empty()).unwrap_or_else(|| {
+            path.file_name()
+                .and_then(|s| s.to_str())
+                .unwrap_or_default()
```
flub, line 532: I know this is existing logic, but we're putting empty strings as names if they're not UTF-8.  How about we use the hash as name instead in that case?

src/provider.rs
```diff
@@ -421,7 +421,13 @@ pub struct Data {
 
 #[derive(Debug)]
 pub enum DataSource {
+    /// A blob of data originating from the filesystem. The name of the blob is derived from
+    /// the filename.
     File(PathBuf),
+    /// NamedFile is treated the same as [`DataSource::File`], except you can pass in a custom name
+    /// for the blob. If `name` is empty, it will derive the blob name from the `path`, in the same
+    /// way that we derive the blob name for [`DataSource::File`].
+    NamedFile { path: PathBuf, name: String },
```
flub, line 430: I think this is fine.  I could be tempted to add shorter constructors by now though, e.g.
```
impl DataSource {
    pub fn new(path: PathBuf) -> Self { todo!() }
    pub fn with_name(path: PathBuf, name: String) -> Self { todo!() }
}
```
but that's optional really.

Locally I already have an `impl From<PathBuf> for DataSource`, (and one for `&Path`) because it saved me typing.  These could also be worth it.

And lastly, truly entering bikeshedding terrain, I wouldn't mind if this was renamed `BlobSource`.  Though I don't mind `DataSource` that much either, it would seem a little more consistent to me.

flub (approved):

src/provider.rs
flub: Ok, I no longer think this is a good idea: the receiver already has the hash when it gets this blob and it can itself figure out what to do if the name is empty, like replace it with a hash

ramfox: Thank you for bringing attention to this though, I forgot we handled the empty string case already so I don't need to add the filter.

src/provider.rs
ramfox: Only thing I didn't do is the name change!

flub (approved):

PR #98, opened 2023-02-02
enforce `MAX_MESSAGE_SIZE` on read

### LINKED ISSUE #20: client & server: set a maximum message size for incoming requests & responses

### LINKED ISSUE #756: client & server: set a maximum message size for incoming requests & responses

ramfox:
closes n0-computer/iroh#756

### DIFF
--- a/src/get.rs
+++ b/src/get.rs
@@ -16,10 +16,9 @@ use crate::bao_slice_decoder::AsyncSliceDecoder;
 use crate::blobs::Collection;
 use crate::protocol::{
     read_bao_encoded, read_lp_data, write_lp, AuthToken, Handshake, Request, Res, Response,
+    MAX_MESSAGE_SIZE,
 };
 use crate::tls::{self, Keypair, PeerId};
-
-const MAX_DATA_SIZE: u64 = 1024 * 1024 * 1024;
 
 #[derive(Clone, Debug)]
 pub struct Options {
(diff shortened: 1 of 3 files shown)


src/get.rs
```diff
@@ -129,10 +128,10 @@ where
                     // server is sending over a collection of blobs
                     Res::FoundCollection { total_blobs_size } => {
                         ensure!(
-                            total_blobs_size <= MAX_DATA_SIZE,
+                            total_blobs_size <= MAX_MESSAGE_SIZE,
                             "size too large: {} > {}",
                             total_blobs_size,
-                            MAX_DATA_SIZE
+                            MAX_MESSAGE_SIZE
```
ramfox, line 134: So... this is probably not what we really want. This is capping the total size of the transfer to `MAX_MESSAGE_SIZE`, rather than capping the specific blob transfers.

However, I believe adding some `MAX_TOTAL_TRANSFER_SIZE` might still be useful. If so, what is a reasonable max for this? a few Gigs? `MAX_MESSAGE_SIZE` is currently 100MiBs

rklaehn:
Added this https://github.com/n0-computer/sendme/pull/99 to allow getting the size out of the AsyncSliceDecoder.

I think the same can be done in bao itself.

src/get.rs
flub: I think a few gigs at least.  What resource are we protecting with `MAX_TOTAL_TRANSFER_SIZE`?  I don't think we should be protecting the user's free disk space.  If we protect the amount of data we need to keep in memory that's something else (but we should fix the protocol to remove those limits).

If the limit is for only protecting the disk space then I'd say we should not have a limit.

dignifiedquire:
@ramfox @rklaehn did n0-computer/sendme#99 fully replace this? can this be closed?

rklaehn:
n0-computer/sendme#99 adds a few things, most notably the ability to get the size from the AsyncSliceDecoder. But it does not limit the `total_blobs_size`. So if we want that it is not a full replacement.

ramfox:
between n0-computer/sendme#99 and n0-computer/sendme#108 this can be closed.

PR #99, opened 2023-02-02
Enforce size limit when transferring blobs

rklaehn:
This adds a new fn read_size to the AsyncSliceReader, and also uses that method to enforce that we don't read oversized messages.

I am a bit confused though. MAX_MESSAGE_SIZE seems to be the maximum size for a protocol message. For that, 100 mb is a bit large. For a file transfer on the other hand, 100mb seems very low. So I can't use sendme to transfer movie files 😞 ...

This PR is using MAX_DATA_SIZE. That is supposed to be the maximum data per item, or in total?

### DIFF
--- a/src/bao_slice_decoder.rs
+++ b/src/bao_slice_decoder.rs
@@ -100,6 +100,15 @@ impl SliceIter {
         self.len = len;
     }
 
+    /// get the length of the slice
+    pub fn len(&self) -> Option<u64> {
+        if self.res.is_some() {
+            Some(self.len)
+        } else {
+            None
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


flub:
100Mb certainly seems too low for files.  OTOH for deserialising postcard it could be a bit large.  But also not sure what size-limits the collection will get into with large deltachat backups.

ramfox (approved):

ramfox:
Regardless of what we do with limiting or not limiting the sizes right now, the ability to know the size of the encoded data still seems useful & I think we should merge this and figure out what limits if any we should have outside of this PR

PR #103, opened 2023-02-02
fix(main): remove stderr tty check

### LINKED ISSUE #95: console logging only works if it's a propper tty
Currently if you run `sendme` in a non tty environment you loose the console output. We should be able to modulate if `-o` is specified, it should maintain logs and only fall back to only content output if its not a tty and there is no output specified.

### LINKED ISSUE #725: console logging only works if it's a propper tty
Currently if you run `sendme` in a non tty environment you loose the console output. We should be able to modulate if `-o` is specified, it should maintain logs and only fall back to only content output if its not a tty and there is no output specified.

dignifiedquire:
- always: RUST_LOG -> STDERR
- always: prints -> STDERR
- istty: progressbar -> STDERR

Closes n0-computer/iroh#725

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1808,7 +1808,6 @@ dependencies = [
  "genawaiter",
  "hex",
  "indicatif",
- "is-terminal",
  "postcard",
  "proptest",
  "rand 0.7.3",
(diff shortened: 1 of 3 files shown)


Arqu (approved):
💙

PR #104, opened 2023-02-02
test: use random free port to avoid collisions

### LINKED ISSUE #49: test: randomize ports
Port assignments should be randomized, rather than hard coded, to avoid test failures

### LINKED ISSUE #745: test: randomize ports

dignifiedquire:
Closes n0-computer/iroh#745

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -192,7 +192,7 @@ mod tests {
 
         let (db, collection_hash) = provider::create_collection(files).await?;
 
-        let addr = format!("127.0.0.1:0").parse().unwrap();
+        let addr = "127.0.0.1:0".parse().unwrap();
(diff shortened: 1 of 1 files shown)


flub (changes requested):
I don't think the `portpicker` crate is that good: it is inherently racy.

We can use the OS-provided mechanism of binding the provider to port `0` and use `Provider::listen_address` to configure the get side in tests.  No new dependencies needed.

dignifiedquire:
My experience when using `0` for doing cross platform things have been very mixed and the result of instability on non linux platforms, so I am not fully convinced it is a good idea to use it :/

ramfox (approved):
lock file conflict?

dignifiedquire:
@flub your solution is simpler, so going with that for now

flub (approved):

PR #106, opened 2023-02-02
feat: switch printing from hex to base64

### LINKED ISSUE #91: Don't use hex for display purposes
We use hex to display several binary things: hashes, auth token, peer ID, ticket.  This is pretty verbose and we can probably use something better.

### LINKED ISSUE #728: Don't use hex for display purposes
We use hex to display several binary things: hashes, auth token, peer ID, ticket.  This is pretty verbose and we can probably use something better.

dignifiedquire:
Why base64? 
Because most of the world uses it already, and the url safe version can be at least easily used in URLs. Also it is very fast :) 

Screenshot 2023-02-02 at 18 27 41

Closes n0-computer/iroh#728

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ rust-version = "1.63"
 [dependencies]
 anyhow = { version = "1", features = ["backtrace"] }
 bao = "0.12.1"
+base64 = "0.21.0"
 blake3 = "1.3.3"
 bytes = "1"
 clap = { version = "4", features = ["derive"] }
@@ -21,7 +22,6 @@ der = { version = "0.6", features = ["alloc", "derive"] }
 ed25519-dalek = { version = "1.0.1", features = ["serde"] }
 futures = "0.3.25"
 genawaiter = { version = "0.99", features = ["futures03"] }
-hex = "0.4.3"
 indicatif = { version = "0.17", features = ["tokio"] }
 postcard = { version = "1", default-features = false, features = ["alloc", "use-std", "experimental-derive"] }
 rand = "0.7"
@@ -43,6 +43,7 @@ zeroize = "1.5"
 
 
 [dev-dependencies]
+hex = "0.4.3"
 proptest = "1.0.0"
 rand = "0.7"
 testdir = "0.7.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,6 @@ console = "0.15.5"
 der = { version = "0.6", features = ["alloc", "derive"] }
 ed25519-dalek = { version = "1.0.1", features = ["serde"] }
 futures = "0.3.25"
-genawaiter = { version = "0.99", features = ["futures03"] }
 indicatif = { version = "0.17", features = ["tokio"] }
 postcard = { version = "1", default-features = false, features = ["alloc", "use-std", "experimental-derive"] }
 rand = "0.7"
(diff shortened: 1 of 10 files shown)


src/bao_slice_decoder.rs
```diff
@@ -806,8 +808,8 @@ mod tests {
                     let data = &slice[offset..offset + 64];
                     let used = |b| if b { "*" } else { " " };
                     println!("hashes root={is_root}");
-                    println!("{} {}", hex::encode(&data[..32]), used(left));
-                    println!("{} {}", hex::encode(&data[32..]), used(right));
+                    println!("{} {}", util::encode(&data[..32]), used(left));
```
rklaehn: I think for the utils for the slice extractor you really want hex.

This is in test, so we can just have hex as a test dependency.

src/bao_slice_decoder.rs
```diff
@@ -818,7 +820,7 @@ mod tests {
                     let data = &slice[offset..offset + size as usize];
                     println!("data range={}..{} root={}", start, start + size, is_root);
                     for chunk in data.chunks(32) {
-                        println!("{}", hex::encode(chunk));
+                        println!("{}", util::encode(chunk));
```
rklaehn: same

rklaehn (approved):
Works for me, but I would prefer if we would use base64 just for the combined token that is going to end up on the QR code.

For a 32 byte hash or peer id, I think hex is fine.

For the debug tools for AsyncSliceDecoder, it *has* to be hex though. (That thing will vanish from the crate as long as we make progress with bao one way or another.. )

src/bao_slice_decoder.rs
dignifiedquire: sure, works for me

dignifiedquire:
> For a 32 byte hash or peer id, I think hex is fine.

I agree in general, but I am really not a big fan in mixing encodings in user facing pieces.

src/bao_slice_decoder.rs
dignifiedquire: fixed

src/bao_slice_decoder.rs
dignifiedquire: fixed

rklaehn:
> > For a 32 byte hash or peer id, I think hex is fine.
> 
> I agree in general, but I am really not a big fan in mixing encodings in user facing pieces.

As long as I don't get weird base64 in my bao format dumper, I am fine with that...

Cargo.toml
```diff
@@ -8,20 +8,19 @@ authors = ["dignifiedquire <me@dignifiedquire.com>"]
 repository = "https://github.com/n0-computer/sendme"
 
 # Sadly this also needs to be updated in .github/workflows/ci.yml
-rust-version = "1.63"
+rust-version = "1.64"
```
flub, line 11: Hum, that's unfortunate as it brings us out of sync with deltachat.  Can this not be avoided?  IIRC the `data-encoding` crate is also popular (and some argue better maintained).

dignifiedquire: this has nothing to do with base64, this is coming from `clap`

dignifiedquire: (I ran `cargo update`..)

flub: Is it that important to have the latest clap in the lock file?  I'm not sure that's worth the tradeoff.  For a library we'll have to get used to providing some kind of msrv for a while.

dignifiedquire: No, but it will happen sooner or later, having to pin to an old version of clap is quite unfortunate.

flub:
(don't consider this PR blocked because of my msrv musings btw - not sure how important they are)

PR #107, opened 2023-02-02
feat(provider): graceful shutdown

### LINKED ISSUE #77: Graceful shutdown of provider
Provider should be able to be requested to be shut down.  At this point it should stop accepting new connections but continue any current transfer.  A timeout for continuing the existing transfer is not needed, if that takes to long a hard abort by the user is fine.

### LINKED ISSUE #736: Graceful shutdown of node
Provider should be able to be requested to be shut down.  At this point it should stop accepting new connections but continue any current transfer.  A timeout for continuing the existing transfer is not needed, if that takes to long a hard abort by the user is fine.

dignifiedquire:
This stops accepting new connections, but does not currently abort transfers.

Closes n0-computer/iroh#736

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,7 +9,7 @@ use indicatif::{
 use sendme::protocol::AuthToken;
 use sendme::provider::Ticket;
 use tokio::io::AsyncWriteExt;
-use tokio::sync::{mpsc, Mutex};
+use tokio::sync::Mutex;
 use tracing_subscriber::{fmt, prelude::*, EnvFilter};
 
 use sendme::{get, provider, util, Keypair, PeerId};
@@ -205,9 +205,7 @@ async fn main() -> Result<()> {
                 .println(format!("All-in-one ticket: {}", provider.ticket(hash)))
                 .await;
 
-            let (s, mut r) = mpsc::channel(1);
-            ctrlc::set_handler(move || s.try_send(()).expect("failed to send shutdown signal"))?;
-            r.recv().await;
+            tokio::signal::ctrl_c().await?;
             out_writer.println("Shutting down...").await;
             provider.shutdown().await?;
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,7 +9,7 @@ use indicatif::{
 use sendme::protocol::AuthToken;
 use sendme::provider::Ticket;
 use tokio::io::AsyncWriteExt;
-use tokio::sync::Mutex;
+use tokio::sync::{mpsc, Mutex};
 use tracing_subscriber::{fmt, prelude::*, EnvFilter};
 
 use sendme::{get, provider, util, Keypair, PeerId};
@@ -204,7 +204,12 @@ async fn main() -> Result<()> {
             out_writer
                 .println(format!("All-in-one ticket: {}", provider.ticket(hash)))
                 .await;
-            provider.join().await?;
+
+            let (s, mut r) = mpsc::channel(1);
+            ctrlc::set_handler(move || s.try_send(()).expect("failed to send shutdown signal"))?;
+            r.recv().await;
+            out_writer.println("Shutting down...").await;
+            provider.shutdown().await?;
 
             // Drop tempath to signal it can be destroyed
             drop(tmp_path);
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -5,15 +5,17 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use std::{collections::HashMap, sync::Arc};
 
-use anyhow::{anyhow, bail, ensure, Context, Result};
+use anyhow::{anyhow, bail, ensure, Context as _, Result};
 use bao::encode::SliceExtractor;
 use bytes::{Bytes, BytesMut};
+use derivative::Derivative;
 use s2n_quic::stream::BidirectionalStream;
-use s2n_quic::Server as QuicServer;
+use s2n_quic::{Connection, Server as QuicServer};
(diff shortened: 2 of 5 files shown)


src/main.rs
```diff
@@ -204,7 +204,12 @@ async fn main() -> Result<()> {
             out_writer
                 .println(format!("All-in-one ticket: {}", provider.ticket(hash)))
                 .await;
-            provider.join().await?;
+
+            let (s, mut r) = mpsc::channel(1);
+            ctrlc::set_handler(move || s.try_send(()).expect("failed to send shutdown signal"))?;
```
flub: IIRC the ctrlc crate creates a new thread for the signal handler.  Tokio has `tokio::signal::ctrl_c` which I think avoids this.  Could be worth using that instead?

src/provider.rs
```diff
@@ -121,28 +129,36 @@ impl Builder {
...
-            let db = db.clone();
-            let events = events.clone();
-            tokio::spawn(async move {
-                debug!("connection accepted from {:?}", connection.remote_addr());
-                while let Ok(Some(stream)) = connection.accept_bidirectional_stream().await {
-                    let _ = events.send(Event::ClientConnected {
-                        connection_id: connection.id(),
-                    });
+        loop {
+            tokio::select! {
```
flub, line 134: It's almost always desirable to use the `biased;` argument to `tokio::select!`: it avoids the randdom number generator overhead on every loop and makes it deterministic to reason about the order the branches are executed.  I don't think it makes a serious difference here, but I would used biased and put the shutdown branch first.

flub (commented):
The interesting thing is that we spawn `handle_stream` for each connection and on shutdown only wait for the *server* task to finish, not for all the handle-stream tasks to finish.  So the `Prover::shutdown` method will return before all clients are done and we have no way of knowing when the clients are done.

Could we introduce some synchronisation between all the spawned tasks and the server so we can allow waiting till all is done?

src/main.rs
dignifiedquire: ohh this is nice, I didn't know about it

src/provider.rs
dignifiedquire: yeah, makes sense

dignifiedquire:
@flub added a version that now fully takes care of the spawned tasks. Not the cleanest, but it should do the trick.

src/provider.rs
```diff
@@ -102,47 +104,88 @@ impl Builder {
...
         db: Database,
         token: AuthToken,
         events: broadcast::Sender<Event>,
     ) {
         debug!("\nlistening at: {:#?}", server.local_addr().unwrap());
+        let (mut current_ctx, _handle) = Context::with_parent(&ctx, None);
+
+        loop {
+            tokio::select! {
+                _ = current_ctx.done() => {
```
flub, line 135: Why is this a new child context rather than calling `ctx.done()` directly?

src/provider.rs
```diff
@@ -102,47 +104,88 @@ impl Builder {
...
+                Some(connection) = server.accept() => {
                     let db = db.clone();
                     let events = events.clone();
-                    tokio::spawn(async move {
-                        if let Err(err) = handle_stream(db, token, stream, events).await {
-                            warn!("error: {:#?}", err);
-                        }
-                        debug!("disconnected");
-                    });
+                    let (current_ctx, _handle) = RefContext::with_parent(&ctx, None);
```
flub, line 142: doesn't handle get dropped at the next loop iteration?  And doesn't dropping the handle cancel the context?

I'm still trying to wrap my head around this context thing...

src/provider.rs
```diff
@@ -102,47 +104,88 @@ impl Builder {
...
+
+async fn handle_connection(
+    ctx: RefContext,
+    mut connection: Connection,
+    db: Database,
+    token: AuthToken,
+    events: broadcast::Sender<Event>,
+) {
+    debug!("connection accepted from {:?}", connection.remote_addr());
+    let (mut current_ctx, _handle) = Context::with_parent(&ctx, None);
```
flub, line 158: I'm also unsure why this needs a new child context

src/provider.rs
```diff
@@ -102,47 +104,88 @@ impl Builder {
...
+                let (mut current_ctx, _handle) = Context::with_parent(&ctx, None);
+
+                tokio::spawn(async move {
+                    tokio::select! {
+                        biased;
+
+                        _ = current_ctx.done() => {
+                            return;
+                        }
+                        res = handle_stream(db, token, stream, events) => {
```
flub, line 181: This is still not hooked up to anything, is it not this that needs to signal to the `Provider` instance it originates from when it is done?  Otherwise we're still interrupting mid-transfer I think.

flub (commented):
I have to admit I'm having a hard time wrapping my head around the tokio-context stuff.  It took me forever to review this and I'm still not really sure I understood things correctly.  I think I can see how it's useful to tell the connection to stop accepting new streams though.  But I don't understand why it builds this hierarchy.

src/provider.rs
dignifiedquire: because `RefContext`s `done` isn't `Send` making it not work here :/

src/provider.rs
dignifiedquire: maybe..

src/provider.rs
dignifiedquire: Currently this will be aborted in the middle of the transfer, yes, which I thought was the goal

dignifiedquire:
Lets not merge this yet, I am not happy with the context setup atm

PR #110, opened 2023-02-03
isolate binary only dependencies behind a feature flag

### LINKED ISSUE #74: Build binary only as optional feature
Right now crates which want to depend on this as a library pull in all sorts of dependencies only for the binary, clap, indicatif etc.  It would be nice to hide these behind an optional cargo feature so that the dependencies are minimised when using it as a library.

### LINKED ISSUE #738: Build binary only as optional feature
Right now crates which want to depend on this as a library pull in all sorts of dependencies only for the binary, clap, indicatif etc.  It would be nice to hide these behind an optional cargo feature so that the dependencies are minimised when using it as a library.

ramfox:
closes n0-computer/iroh#738 

@flub thoughts? Is there a better way to structure this?
My only hesitation is that this makes using the cli via `cargo run` more cumbersome since we now always have to add the `--features="cli"` flag:
`cargo run --features="cli" -- provide FILE`
`cargo run --features="cli" -- get-ticket TICKET -o new_dir`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,9 +16,12 @@ bao = "0.12.1"
 base64 = "0.21.0"
 blake3 = "1.3.3"
 bytes = "1"
+clap = { version = "4", features = ["derive"], optional = true }
+console = { version = "0.15.5", optional = true }
 der = { version = "0.6", features = ["alloc", "derive"] }
 ed25519-dalek = { version = "1.0.1", features = ["serde"] }
 futures = "0.3.25"
+indicatif = { version = "0.17", features = ["tokio"], optional = true }
 postcard = { version = "1", default-features = false, features = ["alloc", "use-std", "experimental-derive"] }
 rand = "0.7"
 rcgen = "0.10"
@@ -37,10 +40,6 @@ webpki = "0.22"
 x509-parser = "0.14"
 zeroize = "1.5"
 
-clap = { version = "4", features = ["derive"], optional = true }
-console = { version = "0.15.5", optional = true }
-indicatif = { version = "0.17", features = ["tokio"], optional = true }
-
 [dev-dependencies]
 hex = "0.4.3"
 proptest = "1.0.0"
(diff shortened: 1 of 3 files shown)


rklaehn (approved):
Nice and clean.

I actually find cargo somewhat nice to use...
https://news.ycombinator.com/item?id=34577844

dignifiedquire (approved):

Cargo.toml
```diff
@@ -41,8 +38,19 @@ webpki = "0.22"
 x509-parser = "0.14"
 zeroize = "1.5"
 
+clap = { version = "4", features = ["derive"], optional = true }
+console = { version = "0.15.5", optional = true }
+indicatif = { version = "0.17", features = ["tokio"], optional = true }
+is-terminal = { version = "0.4", optional = true }
```
flub: nitpicking, but i wouldn't group these separately because they're optional.  now `cargo-add` will think the list is not ordered and no longer order, putting things at the bottom even though they're not optional.

flub (approved):
nice that this doesn't even require any code change!

flub:
We could put `cli` in the default features if you want to get rid of the `cargo run --all-features` thing.  Though I guess it depends on whether the cli or the lib is the primary usecase.

dignifiedquire:
> We could put `cli` in the default features if you want to get rid of the `cargo run --all-features` thing. Though I guess it depends on whether the cli or the lib is the primary usecase.

It would be good if running `cargo install sendme` just works, not entirely sure how that is handled

flub:
> > We could put `cli` in the default features if you want to get rid of the `cargo run --all-features` thing. Though I guess it depends on whether the cli or the lib is the primary usecase.
> 
> It would be good if running `cargo install sendme` just works, not entirely sure how that is handled

I think that uses the default features by default, `cargo-install` has `--no-default-features`, `--all-features` and `--features=...` flags just like expected.

I'm fine with putting cli in the default features, but maybe call it out in the readme that you should use `no_default_features` if you're a lib.

dignifiedquire:
> I'm fine with putting cli in the default features, but maybe call it out in the readme that you should use `no_default_features` if you're a lib.

sgtm

PR #112, opened 2023-02-03
feat: switch from s2n-quic to quinn

dignifiedquire:
Tests are failing because of port reuse issues and shutdown, should be fixed with the other PRs merged

Needed to allow work on hole punching in the short term.

Ref n0-computer/iroh#722 

See

- https://github.com/aws/s2n-quic/issues/1607
- https://github.com/aws/s2n-quic/issues/1608

### DIFF
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -17,8 +17,6 @@ use std::{collections::HashMap, sync::Arc};
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use bao::encode::SliceExtractor;
 use bytes::{Bytes, BytesMut};
-use s2n_quic::stream::BidirectionalStream;
-use s2n_quic::Server as QuicServer;
 use serde::{Deserialize, Serialize};
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 use tokio::sync::broadcast;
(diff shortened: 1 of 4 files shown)


rklaehn:
I would say we do this so we can make some progress with hole punching etc, but try to keep the door open to moving back to s2n_quic...

src/provider.rs
```diff
@@ -245,12 +250,10 @@ impl Provider {
...
     db: Database,
     token: AuthToken,
-    stream: BidirectionalStream,
+    connection_id: u64,
+    (mut writer, mut reader): (quinn::SendStream, quinn::RecvStream),
     events: broadcast::Sender<Event>,
 ) -> Result<()> {
-    debug!("stream opened from {:?}", stream.connection().remote_addr());
-    let connection_id = stream.connection().id();
-    let (mut reader, mut writer) = stream.split();
```
rklaehn, line 283: This seems to be one of the biggest differences - s2n_quic gives you a single endpoint that you can split, quinn gives you the two ends pre-split. I wonder what the semantics are when you e.g. drop the reader and not the writer. 

Probably best to keep the two together...

rklaehn (approved):

flub (approved):

PR #113, opened 2023-02-03
refactor: make AsyncSliceDecoder pub(crate)

rklaehn:
The public interface is now BaoDecoder. A benefit of this is that we
- hide the exact way we decode bao (bao crate or our own thing)
- hide the s2n_quic::stream::ReceiveStream

### DIFF
--- a/src/get.rs
+++ b/src/get.rs
@@ -67,13 +67,16 @@ pub struct Stats {
     pub mbits: f64,
 }
 
+/// A verified stream of data coming from the provider
+///
+/// We guarantee that the data is correct by incrementally verifying a hash
 #[repr(transparent)]
 #[derive(Debug)]
-pub struct BaoDecoder(AsyncSliceDecoder<ReceiveStream>);
+pub struct DataStream(AsyncSliceDecoder<ReceiveStream>);
 
-impl BaoDecoder {
+impl DataStream {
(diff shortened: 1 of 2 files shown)


src/get.rs
```diff
@@ -66,6 +67,34 @@ pub struct Stats {
     pub mbits: f64,
 }
 
+#[repr(transparent)]
```
flub, line 73: Is this repr important?  I thought this only really mattered if you're doing FFI stuff.  The compiler should not use extra space for a tuple struct anyway I thought?

src/get.rs
```diff
@@ -66,6 +67,34 @@ pub struct Stats {
...
+#[derive(Debug)]
+pub struct BaoDecoder(AsyncSliceDecoder<ReceiveStream>);
```
flub: nit: If it's pub it should have a nice doc comment

flub (approved):

src/get.rs
dignifiedquire: nit: I would name it sth else, just `DataStream` or sth, it's not really relevant for the user that this happens to be decoding data using bao

src/get.rs
rklaehn: No, it should not matter. It is just something I add to newtype structs so I get a compile error when I add something to the struct that makes it no longer a pure newtype wrapper.

PR #115, opened 2023-02-03
feat: introduce custom Hash type

dignifiedquire:
- avoids relying on blake3::Hash as a type in the public api
- ensures consistent encoding and decoding
- fixes bug where hex decoding was expected in the UI


Arqu (approved):
LGTM, this is a very healthy change in general.

ramfox (approved):

PR #116, opened 2023-02-06
Reduce surface area

rklaehn:
First set of surface area reduction

This is the pretty straightforward part. To reduce surface area further you would have to change the get and provider side api to be more high level.

E.g. get::run with all these callbacks can not really have an extremely low surface area, same for provider::Provider::subscribe with all the internal events.

### DIFF
--- a/src/get.rs
+++ b/src/get.rs
@@ -1,4 +1,8 @@
-//! The client side api
+//! The client side API
+//!
+//! The main entry point is [`run`]. This function takes callbacks that will
+//! be invoked when blobs or collections are received. It is up to the caller
+//! to store the received data.
 use std::fmt::Debug;
 use std::io;
 use std::net::SocketAddr;
@@ -71,8 +75,14 @@ pub struct Stats {
     pub data_len: u64,
     /// The time it took to transfer the data
     pub elapsed: Duration,
+}
+
+impl Stats {
     /// Transfer rate in megabits per second
-    pub mbits: f64,
+    pub fn mbits(&self) -> f64 {
+        let data_len_bit = self.data_len * 8;
+        data_len_bit as f64 / (1000. * 1000.) / self.elapsed.as_secs_f64()
+    }
 }
 
 /// A verified stream of data coming from the provider
@@ -222,15 +232,8 @@ where
                 writer.close().await?;
 
                 let elapsed = now.elapsed();
-                let elapsed_s = elapsed.as_secs_f64();
-                let data_len_bit = data_len * 8;
-                let mbits = data_len_bit as f64 / (1000. * 1000.) / elapsed_s;
 
-                let stats = Stats {
-                    data_len,
-                    elapsed,
-                    mbits,
-                };
+                let stats = Stats { data_len, elapsed };
 
                 Ok(stats)
             }
--- a/src/get.rs
+++ b/src/get.rs
@@ -1,3 +1,4 @@
+//! The client side api
 use std::fmt::Debug;
 use std::io;
 use std::net::SocketAddr;
@@ -24,9 +25,12 @@ pub use crate::util::Hash;
 
 const MAX_DATA_SIZE: u64 = 1024 * 1024 * 1024;
 
+/// Options for the client
 #[derive(Clone, Debug)]
 pub struct Options {
+    /// The address to connect to
     pub addr: SocketAddr,
+    /// The peer id to expect
     pub peer_id: Option<PeerId>,
 }
 
@@ -63,8 +67,11 @@ async fn setup(opts: Options) -> Result<(Client, Connection)> {
 /// Stats about the transfer.
 #[derive(Debug, Clone, PartialEq)]
 pub struct Stats {
+    /// The number of bytes transferred
     pub data_len: u64,
+    /// The time it took to transfer the data
     pub elapsed: Duration,
+    /// Transfer rate in megabits per second
     pub mbits: f64,
 }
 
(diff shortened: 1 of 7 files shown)


dignifiedquire (approved):

src/get.rs
```diff
@@ -63,8 +67,11 @@ async fn setup(opts: Options) -> Result<(Client, Connection)> {
 /// Stats about the transfer.
 #[derive(Debug, Clone, PartialEq)]
 pub struct Stats {
+    /// The number of bytes transferred
     pub data_len: u64,
+    /// The time it took to transfer the data
     pub elapsed: Duration,
+    /// Transfer rate in megabits per second
```
flub, line 81: Probably not for this PR.  But that's an odd unit.  I wonder if there's a better type for this, something like a Duration + number of bytes.  If you want this unit you can then easily enough create it, or even add it as a method.

/cc @Arqu

src/provider.rs
```diff
@@ -1,3 +1,4 @@
+//! Provider api
```
flub: fwiw as part of public docs - that is, if the module even stays public - I would follow this up with a short paragraph describing what the provider is and how it works, from a super high level.  I usually err on the side of being repetitive in docs.

src/provider.rs
```diff
@@ -25,7 +26,15 @@ use crate::util::{self, Hash};
 const MAX_CONNECTIONS: u64 = 1024;
 const MAX_STREAMS: u64 = 10;
 
-pub type Database = Arc<HashMap<Hash, BlobOrCollection>>;
+/// Database containing content-addressed data (blobs or collections).
+#[derive(Debug, Clone)]
+pub struct Database(Arc<HashMap<Hash, BlobOrCollection>>);
```
flub, line 39: :+1:

src/provider.rs
```diff
@@ -25,7 +26,15 @@ use crate::util::{self, Hash};
...
+
+impl Database {
+    fn get(&self, key: &Hash) -> Option<&BlobOrCollection> {
```
flub, line 42: this raises the question of `iter` etc as well.  but this is fine to start with

src/provider.rs
```diff
@@ -168,20 +177,32 @@ pub struct Provider {
 /// Events emitted by the [`Provider`] informing about the current status.
 #[derive(Debug, Clone)]
 pub enum Event {
+    /// A new client connected to the provider.
     ClientConnected {
+        /// The quic connection id.
```
flub: As this is public API I would be hesitant to commit to this being exactly the quic connection id.  I'd be more vague and say "A connection ID unique for the currently connected clients" or so.

src/provider.rs
```diff
@@ -417,20 +438,28 @@ pub struct Data {
...
+    NamedFile {
+        /// Path to the file
+        path: PathBuf,
+        /// Custom name
+        name: String,
+    },
 }
 
 impl DataSource {
+    /// Create a new [`DataSource`] from a [`PathBuf`].
```
flub: very much bikeshedding: but I've always been thought to write this in 3rd party form: `Creates a new [`DataSource`] from a [`PathBuf`].`
(kind of applicable everywhere)

src/provider.rs
```diff
@@ -600,11 +629,13 @@ pub struct Ticket {
 }
 
 impl Ticket {
+    /// Deserializes from bytes.
```
flub, line 641: hey look, 3rd party!  seems like you also do this sometimes...

src/tls.rs
```diff
@@ -22,6 +22,7 @@ use crate::util;
 // TODO: change?
 const P2P_ALPN: [u8; 6] = *b"libp2p";
 
+/// A keypair.
```
flub, line 25: Do we expose this at all or is this entire module internal?  I feel like it should be entirely internal.  If ever not it should probably be its own crate.

Not that that doesn't mean the docstrings don't need to be added, they're always good.

flub (approved):

src/get.rs
rklaehn: Isn't this redundant with the other two fields anyway?

flub: oh, yeah.  how did i totally miss that :smile:

src/provider.rs
rklaehn: To be completely honest, github copilot was not completely uninvolved in the creation of this code...

src/tls.rs
rklaehn: This is needed in the provider builder, so making this private would require some more changes. Probably best left for another PR.

flub: oh, fine.  missed that.

src/provider.rs
```diff
@@ -179,12 +187,12 @@ pub struct Provider {
 pub enum Event {
     /// A new client connected to the provider.
     ClientConnected {
-        /// The quic connection id.
+        /// An unique connection id.
```
flub, line 190:
```suggestion
        /// A unique connection id.
```

src/provider.rs
```diff
@@ -179,12 +187,12 @@ pub struct Provider {
...
         connection_id: u64,
     },
     /// A request was received from a client.
     RequestReceived {
-        /// The quic connection id.
+        /// An unique connection id.
```
flub, line 195:
```suggestion
        /// A unique connection id.
```

src/provider.rs
```diff
@@ -193,7 +201,7 @@ pub enum Event {
     },
     /// A request was completed and the data was sent to the client.
     TransferCompleted {
-        /// The quic connection id.
+        /// An unique connection id.
```
flub, line 204:
```suggestion
        /// A unique connection id.
```

flub (approved):

PR #121, opened 2023-02-06
feat: Allow shutdown of Provider

flub:
This adds Provider::shutdown allowing to stop the Provider without
having to abort it.  On the wire however it still does not do anything
graceful: all the connections are stopped and any in-transit data is
lost.

Additionally this adds tracing spans to the connection handling.

Finally it fixes a bug where read_lp would not return None if the
connection was closed normally.

Part of n0-computer/iroh#736 but this is not very graceful on the wire yet.

### DIFF
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -1,5 +1,6 @@
 //! Protocol for communication between provider and client.
 use std::fmt::Display;
+use std::io;
 use std::str::FromStr;
 
 use anyhow::{ensure, Result};
@@ -84,7 +85,11 @@ pub(crate) async fn read_lp<'a, R: AsyncRead + Unpin, T: Deserialize<'a>>(
     buffer: &'a mut BytesMut,
 ) -> Result<Option<(T, usize)>> {
     // read length prefix
-    let size = read_prefix(&mut reader).await?;
+    let size = match read_prefix(&mut reader).await {
+        Ok(size) => size,
+        Err(err) if err.kind() == io::ErrorKind::UnexpectedEof => return Ok(None),
+        Err(err) => return Err(err.into()),
+    };
     let mut reader = reader.take(size);
 
     let size = usize::try_from(size)?;
@@ -152,7 +157,7 @@ pub(crate) async fn read_lp_data<R: AsyncRead + Unpin>(
     Ok(Some(response))
 }
 
-async fn read_prefix<R: AsyncRead + Unpin>(mut reader: R) -> Result<u64> {
+async fn read_prefix<R: AsyncRead + Unpin>(mut reader: R) -> Result<u64, io::Error> {
     // read length prefix
     let size = reader.read_u64_le().await?;
     Ok(size)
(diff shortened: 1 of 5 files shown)


flub:
Apologies this kind of merges three changes in one PR.  Happy to untangle if preferred.

src/protocol.rs
```diff
@@ -84,7 +85,11 @@ pub(crate) async fn read_lp<'a, R: AsyncRead + Unpin, T: Deserialize<'a>>(
     buffer: &'a mut BytesMut,
 ) -> Result<Option<(T, usize)>> {
     // read length prefix
-    let size = read_prefix(&mut reader).await?;
+    let size = match read_prefix(&mut reader).await {
+        Ok(size) => size,
+        Err(err) if err.kind() == io::ErrorKind::UnexpectedEof => return Ok(None),
```
flub, line 90: The bugfix: this function never returned None otherwise.

src/provider.rs
```diff
@@ -283,6 +285,44 @@ impl Future for Provider {
...
+    let remote_addr = connecting.remote_address();
+    let connection = match connecting.await {
+        Ok(conn) => conn,
+        Err(err) => {
+            warn!(%remote_addr, "Error connecting: {err:#}");
+            return;
+        }
+    };
+    let connection_id = connection.stable_id() as u64;
+    let span = debug_span!("connection", connection_id, %remote_addr);
```
flub, line 303: the two tracing spans introduced in this PR are entirely contained in this function, one for the connection and one for the stream.

src/provider.rs
```diff
@@ -283,6 +285,44 @@ impl Future for Provider {
...
+    async move {
+        while let Ok(stream) = connection.accept_bi().await {
```
rklaehn, line 305: So how is this cancelled? Just by dropping the task?

rklaehn (approved):
Never heard of a CancellationToken. It seems to be made for this, however I don't quite get what the advantage is over a simple onehot sender/recever pair...

src/provider.rs
flub: On line 168 the whole `Endpoint` is closed. That has the effect that any quinn call, like this `.accept_bi` will return an `Err(ConnectionError::LocallyClosed)`.  At which point this while loop stops and the task finishes.  So the tasks are not dropped, they finish naturally with an error.

flub:
> Never heard of a CancellationToken. It seems to be made for this, however I don't quite get what the advantage is over a simple onehot sender/recever pair...

I was originally thinking of using a oneshot.  However I also originally thought I had to terminate a bunch of tasks before I realised I could call `Endpoint::close` (though later when we do this more gracefully maybe we will need to signal multiple tasks again).  I then wondered whether to use a broadcast channel but I've used cancellation tokens of sorts before so knew of the concept and did some searching.  @dignifiedquire used `tokio-context` in n0-computer/sendme#107 but I found that API to be difficult to understand.  And it turned out we already use tokio-util as well, so I went with this, I found it a nice enough API to use.

PR #122, opened 2023-02-07
fix: catch `ctrl-c` to allow a normal shutdown & clean up any tempfiles that are created

### LINKED ISSUE #102: tempfiles are not cleaned up
If you download a single file with output folder `foo`, and cancel the transfer halfway through, the tempfile is inside of the `foo` directory and is not deleted, but stays around.

### LINKED ISSUE #724: tempfiles are not cleaned up
If you download a single file with output folder `foo`, and cancel the transfer halfway through, the tempfile is inside of the `foo` directory and is not deleted, but stays around.

ramfox:
closes n0-computer/iroh#724

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-use std::{fmt, net::SocketAddr, path::PathBuf, str::FromStr, sync::Arc};
+use std::{fmt, net::SocketAddr, path::PathBuf, str::FromStr};
 
 use anyhow::{bail, Context, Result};
 use clap::{Parser, Subcommand};
@@ -215,7 +215,15 @@ async fn main() -> Result<()> {
             }
             let token =
                 AuthToken::from_str(&token).context("Wrong format for authentication token")?;
-            get_interactive(*hash.as_hash(), opts, token, out).await?;
+            tokio::select! {
+                _ = tokio::signal::ctrl_c() => {
+                    println!("Ending transfer early...");
+                    Ok(())
+                }
+                res = get_interactive(*hash.as_hash(), opts, token, out) => {
+                    res
+                }
+            }
         }
         Commands::GetTicket { out, ticket } => {
             let Ticket {
@@ -228,7 +236,15 @@ async fn main() -> Result<()> {
                 addr,
                 peer_id: Some(peer),
             };
-            get_interactive(hash, opts, token, out).await?;
+            tokio::select! {
+                _ = tokio::signal::ctrl_c() => {
+                    println!("Ending transfer early...");
+                    Ok(())
+                }
+                res = get_interactive(hash, opts, token, out) => {
+                    res
+                }
+            }
         }
         Commands::Provide {
             path,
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -194,6 +199,13 @@ async fn main() -> Result<()> {
                 builder = builder.auth_token(auth_token);
             }
             let provider = builder.spawn()?;
+            let cancel_token = provider.cancel();
```
ramfox: couldn't figure out how to use `Provider::shutdown` here without running into ownership issues, so I've made a way to access the `cancel_token` directly.

Please let me know if I've missed something obvious.

It works fine for this current iteration, but when we have more things to clean up, we may want an actual shutdown method. So I'm not sure the best way to handle this.

src/main.rs
```diff
@@ -241,6 +250,23 @@ async fn get_interactive(
...
+            .expect("failed to listen for `ctrl-c`");
+        let clean_up = clean_up_clone.lock().await;
+        for tmp_file in clean_up.iter() {
+            if let Err(e) = tokio::fs::remove_file(tmp_file).await {
+                tracing::warn!("could not remove temporary file {tmp_file:?}: {e}");
+            }
+        }
+        std::process::exit(0);
+    });
+
```
ramfox: TBH, this feels pretty goofy.

src/main.rs
```diff
@@ -328,6 +357,12 @@ async fn get_interactive(
                     Ok::<_, anyhow::Error>((temp_file, dup))
                 })
                 .await??;
+
+                {
+                    let mut clean_up = clean_up.lock().await;
+                    (*clean_up).push(temp_file.path().to_path_buf());
+                }
```
ramfox: Another option would be to spawn a task for each file that potentially needs cleaning up & listen for `ctrl-c` on each of them. Wasn't sure which was less gross.

I'm sure there are other versions that I'm not thinking of.

src/provider.rs
```diff
@@ -268,9 +266,18 @@ impl Provider {
...
+    /// [`Provider`] will complete.
+    ///
+    /// The shutdown behaviour will become more graceful in the future.
+    pub fn shutdown(&self) {
+        self.cancel_token.cancel();
+    }
+
+    /// Gives you a [`CancellationToken`] that you can use to shutdown the provider
+    pub fn cancel(&self) -> CancellationToken {
+        self.cancel_token.clone()
```
ramfox: will add better documentation if this is the route we decide to go down.

flub:
I changed the PR base to `flub/shutdown` as it makes the diff more readable.  github will automatically switch that back to `main` once `flub/shutdown` is merged in main.

src/main.rs
flub: Wait, this is the provider side?  I think we can skirt all these troubles and use `tempfile::tempfile()` instead of a `NamedTempFile` for the temporary file used in the provider.  This will ensure the temp file will be cleaned up by the OS when the process exits and we can punt on ctrl-c handling in the provider to the graceful shutdown ticket.

But to talk about the lifetime issues, `Provider::shutdown` takes `&self` and so do all the other methods apart from the `Future::poll`.  So we might be able to put an `Arc` around the provider to be able to call shutdown.  It would limit being able to await it though, so I'm not entirely sure this would work out.

What I did in the DC PR was create a new task which itself owns the `Provider` instance.  This task's purpose is to supervise the provider and it itself can wait for an async signal, e.g. a channel or cancel token or ctrl-c signal, and since it owns everything can call all methods on the provider.  It's a bit funky and it made me wonder if the shutdown method should really be some kind of handle, say just a oneshot sender.  But really the `Provider` struct is already meant to be the handle and it already is very similar to a tokio `JoinHandle` so I don't know, I think we should try keep `Provider` be the handle.

But really, I think the first paragraph of my response here is the right path.

src/main.rs
```diff
@@ -241,6 +250,23 @@ async fn get_interactive(
...
+            .await
+            .expect("failed to listen for `ctrl-c`");
+        let clean_up = clean_up_clone.lock().await;
+        for tmp_file in clean_up.iter() {
+            if let Err(e) = tokio::fs::remove_file(tmp_file).await {
+                tracing::warn!("could not remove temporary file {tmp_file:?}: {e}");
+            }
+        }
+        std::process::exit(0);
+    });
```
flub: Goofy good or goofy bad?  I'm going with goofy bad as I'm not 100% sure the mutex can never result in a deadlock.

Anyway, I'm pretty sure we don't need to do any of this explicit cleanup or deal with an `Arc`ed mutex.  The whole point of the tempfile crate is that it does this for us.  It sets up the right destructors.  Our problem is that hitting `Ctrl-c` or otherwise interrupting the client process is not handled at all resulting in the OS just killing the process without running the destructors.

What we need to do instead is to hook up ctrl-c so that our rust code exists main normally, doesn't even have to be successful.  Exiting main in our rust code will ensure the tempfile destructors will run and the temp files will be cleaned up.

src/main.rs
flub: Do you think we need to clean up **all** the files of the collection?  I was only thinking of cleaning up the temporary files but once their checksum matches and they are moved to their final destination I don't think I'd bother with trying to clean them up if later on the transfer fails.  I'd only try and clean up the last partially-written tempfile.  Which is done by making sure the tempfile destructors run as explained in my comment above.

If you really think we should clean up **all** files then all I say about tempfile destructors isn't right and your approach is probably closer to the mark than my suggestions.


Responding to the technicality of spawning a task per file or building a list of files.  Tokio tasks are cheap, and you can have many.  But having way, way, way many of them does eventually slow down the executor.  So you're always better off storing your work in messages/data than creating loads of long-running futures.  Generally keeping your number of long-running futures bounded (a task per file I would consider unbounded) and preferably somewhat low-ish is desirable.

src/provider.rs
flub: If all my assumptions above are right I think we can postpone having to provide an API like this.
(In any case, I'd bikeshed for it to be called `.cancel_token()` :upside_down_face: )

flub (commented):
So for get, if my assumptions are right about how it should behave (see inline comments) then I think the way to go is to put the `get::run(...)` line inside a `tokio::select!` while selecting the ctrl-c future together with it.

Technically we don't install the ctrl-c handler right at the start of the process with that approach, but we also don't do anything yet that needs cleaning up.

src/main.rs
ramfox: If we use `tempfile::tempfile`, then we don't have a path to the file to add to `sources`, so the way we have things set up currently (with DataSource) means we need to use `NamedTempFile` as is :( 

This would only spawn a single task (rather than in the unbounded case on the `get` side).

We could potentially change `Data` and `DataSource` to allow a `TempPath` (as well as a `Path`). Then, the tempfile is only dropped when the database itself is dropped. That might be an excessive change at this point in the process to remove a single task, though.

---

On the `Provider` side, I used the DC PR as a model. As you mentioned, wrapping `Provider` in an `Arc` caused drama when trying to `await` the `Provider`.

src/main.rs
ramfox: Yea, bad goofy. 

You are totally correct about just needing to exit `main` normally! massive overkill on my part :)

Thanks so much! This actually streamlines the entire pr.

src/main.rs
ramfox: I was only intending to remove files that were still temporary! 

Thanks for the perspective re: task per unbounded file.

src/main.rs
ramfox: re comment https://github.com/n0-computer/sendme/pull/122/files#r1099680967, my above comment is no longer relevant!

ramfox:
> So for get, if my assumptions are right about how it should behave (see inline comments) then I think the way to go is to put the get::run(...) line inside a tokio::select! while selecting the ctrl-c future together with it.
>
>Technically we don't install the ctrl-c handler right at the start of the process with that approach, but we also don't do anything yet that needs cleaning up.

Okay, so let me know if you think this is overkill. But rather than waiting inside `get_interactive`, we now handle `ctrl-c` in each `Commands` match. 

We could handle this one step "higher" and have each match arm return some `Future Fn` that we then `select` against `ctrl-c`, but this way we can have individualized output for each arm. I'm not married to that.

Also, now that we are guaranteed a normal shutdown for `Commands::Provide`, we could potentially impl `Drop` on `Provider` that calls `Provider::shutdown` & be sure that we clean up the provider gracefully on `ctrl-c`

src/main.rs
```diff
@@ -215,7 +215,15 @@ async fn main() -> Result<()> {
             }
             let token =
                 AuthToken::from_str(&token).context("Wrong format for authentication token")?;
-            get_interactive(*hash.as_hash(), opts, token, out).await?;
+            tokio::select! {
```
flub, line 218: This is an oddity of the `tokio::select!` API, but it is almost always better to use the biased form since it avoids generating a random number on each call (not that this is in a loop so it doesn't matter much).  It also makes it easier to reason about which branch will be preferred.  For example here I'd prefer the get_interactive branch first if both ctrl-c and get_interactive complete at the same time.  That is the opposite order than they are written now.
```suggestion
            tokio::select! {
                biased;
```

src/main.rs
```diff
@@ -228,80 +236,104 @@ async fn main() -> Result<()> {
...
+        }
+    }
+}
+
+async fn provide_interactive(
+    path: Option<PathBuf>,
+    addr: Option<SocketAddr>,
+    auth_token: Option<String>,
+    key: Option<PathBuf>,
+) -> Result<()> {
```
flub, line 276: I'm assuming this is entirely moved code with no changes.  That kind of thing is still too hard to review with our current tools, so if not please point at the important bits.

flub (approved):
Nice, I think this is a very reasonable way to go about it.

src/main.rs
ramfox: Yes, it is entirely moved with no changes.

That's a good point re: noting that there are no changes when making a move like this, thank you.

src/main.rs
ramfox: ahhh yes, I learned this while reviewing one of your PRs & I forgot to incorporate it. Thank you.

flub (approved):

PR #123, opened 2023-02-07
Add newtype wrapper for Hash that is compatible with Cid

rklaehn:
Implements n0-computer/iroh#762

Spent some effort making the fast path to_string and parse 0 alloc by directly using the data-encoding crate. Not sure if we need the multibase crate, but to be really compatible with cid I guess we do.

So what I don't know is where I am supposed to use this. Replace to_string etc. of sendme::util::Hash? Or just in selected places?

I put it all in main for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,10 +18,12 @@ blake3 = "1.3.3"
 bytes = "1"
 clap = { version = "4", features = ["derive"], optional = true }
 console = { version = "0.15.5", optional = true }
+data-encoding = "2.3.3"
 der = { version = "0.6", features = ["alloc", "derive"] }
 ed25519-dalek = { version = "1.0.1", features = ["serde"] }
 futures = "0.3.25"
 indicatif = { version = "0.17", features = ["tokio"], optional = true }
+multibase = "0.9.1"
 postcard = { version = "1", default-features = false, features = ["alloc", "use-std", "experimental-derive"] }
 quinn = "0.9.3"
 rand = "0.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,12 +18,12 @@ blake3 = "1.3.3"
 bytes = "1"
 clap = { version = "4", features = ["derive"], optional = true }
 console = { version = "0.15.5", optional = true }
-data-encoding = "2.3.3"
+data-encoding = { version = "2.3.3", optional = true }
 der = { version = "0.6", features = ["alloc", "derive"] }
 ed25519-dalek = { version = "1.0.1", features = ["serde"] }
 futures = "0.3.25"
 indicatif = { version = "0.17", features = ["tokio"], optional = true }
-multibase = "0.9.1"
+multibase = { version = "0.9.1", optional = true }
 postcard = { version = "1", default-features = false, features = ["alloc", "use-std", "experimental-derive"] }
 quinn = "0.9.3"
 rand = "0.7"
@@ -50,7 +50,7 @@ testdir = "0.7.1"
 
 [features]
 default = ["cli"]
-cli = ["clap", "console", "indicatif"]
+cli = ["clap", "console", "indicatif", "data-encoding", "multibase"]
 
 [[bin]]
 name = "sendme"
--- a/src/main.rs
+++ b/src/main.rs
@@ -103,7 +103,7 @@ impl OutWriter {
 
 #[repr(transparent)]
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct Blake3Cid(Hash);
+struct Blake3Cid(Hash);
 
 const CID_PREFIX: [u8; 4] = [
     0x01, // version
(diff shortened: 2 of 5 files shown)


src/main.rs
```diff
@@ -101,13 +101,93 @@ impl OutWriter {
...
+        );
+        anyhow::ensure!(bytes[0..4] == CID_PREFIX, "invalid cid prefix");
+        let mut hash = [0u8; 32];
+        hash.copy_from_slice(&bytes[4..36]);
+        Ok(Blake3Cid(Hash::from(hash)))
+    }
+}
+
+impl fmt::Display for Blake3Cid {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
rklaehn, line 146: If main is all we use this thing for, this is probably a bit overdone.

It is zero alloc, which is refreshing compared to the usual multi... stuff.

src/provider.rs
```diff
@@ -595,13 +606,6 @@ pub async fn create_collection(data_sources: Vec<DataSource>) -> Result<(Databas
     let data = postcard::to_slice(&c, &mut buffer)?;
     let (outboard, hash) = bao::encode::outboard(&data);
     let hash = Hash::from(hash);
-    println!("Collection: {hash}\n");
```
rklaehn, line 638: This really had no business being here. Now the printing is done in main.

Cargo.toml
```diff
@@ -18,10 +18,12 @@ blake3 = "1.3.3"
 bytes = "1"
 clap = { version = "4", features = ["derive"], optional = true }
 console = { version = "0.15.5", optional = true }
+data-encoding = { version = "2.3.3", optional = true }
```
flub, line 21: Can't this also do base64?  Can we get rid of the `base64` dep at the same time?

src/main.rs
```diff
@@ -101,13 +101,93 @@ impl OutWriter {
     }
 }
 
+#[repr(transparent)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Blake3Cid(Hash);
+
+const CID_PREFIX: [u8; 4] = [0x01, 0x55, 0x1e, 0x20];
```
dignifiedquire: please document what these mean

src/main.rs
```diff
@@ -101,13 +101,93 @@ impl OutWriter {
     }
 }
 
+#[repr(transparent)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Blake3Cid(Hash);
```
flub: you're in main.rs.  `pub` isn't going to do much is it?

src/main.rs
```diff
@@ -101,13 +101,93 @@ impl OutWriter {
...
+
+const CID_PREFIX: [u8; 4] = [0x01, 0x55, 0x1e, 0x20];
+
+impl Blake3Cid {
+    pub fn new(hash: Hash) -> Self {
+        Blake3Cid(hash)
+    }
+
+    pub fn hash(&self) -> &Hash {
```
flub: maybe `as_hash` to suggest it's a zero-cost retrieve by reference?

src/main.rs
```diff
@@ -101,13 +101,93 @@ impl OutWriter {
...
+        &self.0
+    }
+
+    pub fn into_bytes(&self) -> [u8; 36] {
```
flub: does not consume self, so better to call it `to_bytes` i think?

src/main.rs
rklaehn: Probably a good idea, since I really got no idea. Just replicated what the cid crate does...

I guess "some bytes to make the output compatible to cids" is not a good documentation... 🙄

src/provider.rs
flub: good catch

Cargo.toml
rklaehn: of course it can. It seems to be reasonably well designed.

src/main.rs
rklaehn: 🤷‍♂️ also does not hurt...

src/main.rs
rklaehn: Was called something else before. as_bytes, I think. Clippy complained... Will try to_bytes.

Cargo.toml
rklaehn: Hm, but then it would have to be a non optional dependency.

src/main.rs
rklaehn: No, it also complains about to_bytes.

Cargo.toml
flub: but another required dep would be gone.  anyway, not very important and probably best done separately

src/main.rs
```diff
@@ -101,13 +101,98 @@ impl OutWriter {
...
+impl Blake3Cid {
+    pub fn new(hash: Hash) -> Self {
+        Blake3Cid(hash)
+    }
+
+    pub fn as_hash(&self) -> &Hash {
+        &self.0
+    }
+
+    pub fn as_bytes(&self) -> [u8; 36] {
```
flub, line 124: To me `as_bytes` implies i will get `&[u8]`.  Also this **does** do a conversion, and the [rust naming convention](https://rust-lang.github.io/api-guidelines/naming.html) says that `as_` functions should return something borrowed and be 0-cost.  So I still think `to_bytes` is the right name here.

flub (approved):
apart from one more bit of bikeshedding LGTM

src/main.rs
rklaehn: It seems that to_bytes would be the right name. But clippy complains. Maybe call it `ljhgadlijf` so nobody can complain that it does not fit some naming convention?

PR #124, opened 2023-02-07
feat: Do not error when the provider closes the connection

flub:
This makes it so that the provider is allowed to close the connection
after a transfer is completed, without the getter resulting in an
error.

It modifies the provider to only emit the transfer completed once the
data has actually reached the client.

### DIFF
--- a/src/get.rs
+++ b/src/get.rs
@@ -158,6 +158,8 @@ where
         let used = postcard::to_slice(&req, &mut out_buffer)?;
         write_lp(&mut writer, used).await?;
     }
+    writer.finish().await?;
+    drop(writer);
 
     // 3. Read response
     {
@@ -224,11 +226,7 @@ where
                 }
 
                 // Shut down the stream
-                debug!("shutting down stream");
-                writer.finish().await.or_else(|err| match err {
-                    quinn::WriteError::Stopped(_) | quinn::WriteError::ConnectionLost(_) => Ok(()),
-                    _ => Err(err),
-                })?;
+                drop(reader);
 
                 let elapsed = now.elapsed();
 
@@ -237,7 +235,7 @@ where
                 Ok(stats)
             }
             None => {
-                bail!("provider disconnected");
+                bail!("provider closed stream");
             }
         }
     }
--- a/src/get.rs
+++ b/src/get.rs
@@ -13,7 +13,7 @@ use anyhow::{anyhow, bail, ensure, Result};
 use bytes::BytesMut;
 use futures::Future;
 use postcard::experimental::max_size::MaxSize;
-use tokio::io::{AsyncRead, AsyncWriteExt, ReadBuf};
+use tokio::io::{AsyncRead, ReadBuf};
 use tracing::debug;
 
 use crate::bao_slice_decoder::AsyncSliceDecoder;
@@ -116,7 +116,7 @@ impl AsyncRead for DataStream {
 /// Get a collection and all its blobs from a provider
 pub async fn run<A, B, C, FutA, FutB, FutC>(
     hash: Hash,
-    token: AuthToken,
+    auth_token: AuthToken,
     opts: Options,
     on_connected: A,
     mut on_collection: B,
@@ -145,7 +145,7 @@ where
     // 1. Send Handshake
     {
         debug!("sending handshake");
-        let handshake = Handshake::new(token);
+        let handshake = Handshake::new(auth_token);
         let used = postcard::to_slice(&handshake, &mut out_buffer)?;
         write_lp(&mut writer, used).await?;
     }
@@ -225,7 +225,10 @@ where
 
                 // Shut down the stream
                 debug!("shutting down stream");
-                writer.shutdown().await?;
+                writer.finish().await.or_else(|err| match err {
+                    quinn::WriteError::Stopped(_) | quinn::WriteError::ConnectionLost(_) => Ok(()),
+                    _ => Err(err),
+                })?;
 
                 let elapsed = now.elapsed();
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,14 +21,19 @@ mod tests {
         sync::{atomic::AtomicUsize, Arc},
     };
 
-    use crate::tls::PeerId;
-    use crate::{protocol::AuthToken, util::Hash};
-
-    use super::*;
-    use anyhow::Result;
+    use anyhow::{anyhow, Context, Result};
     use rand::RngCore;
     use testdir::testdir;
-    use tokio::io::AsyncReadExt;
+    use tokio::fs;
+    use tokio::io::{self, AsyncReadExt};
+
+    use crate::{protocol::AuthToken, provider::Event, util::Hash};
+    use crate::{
+        provider::{create_collection, Provider},
+        tls::PeerId,
+    };
+
+    use super::*;
 
     #[tokio::test]
     async fn basics() -> Result<()> {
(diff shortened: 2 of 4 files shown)


flub:
Ugh, this is **wrong**.

If the provider closes all the connections this brutally then there is no guarantee that the getter really receives all the data in time.

Guess we really do need graceful termination of the sendme provider.

flub:
> Ugh, this is **wrong**.
> 
> If the provider closes all the connections this brutally then there is no guarantee that the getter really receives all the data in time.
> 
> Guess we really do need graceful termination of the sendme provider.

Turns out the fix was a lot easier than doing all of graceful termination: https://github.com/n0-computer/sendme/pull/124/commits/3d3e055d7cc02f2d42a4d4793d53eb529d9e6cda

PTAL

flub:
Ugh, realised that this already changes the protocol to one-request-per-stream because `SendStream::finish` does make it so the provider can no longer send anything.  So this is **still** incomplete: the stream should be closed entirely after handling a single request

src/get.rs
```diff
@@ -158,6 +158,8 @@ where
         let used = postcard::to_slice(&req, &mut out_buffer)?;
         write_lp(&mut writer, used).await?;
     }
+    writer.finish().await?;
+    drop(writer);
```
flub, line 162: This isn't strictly needed, I'm open to remove it again.  Currently the provider tells the getter to stop already, this is the symmetric part enforcing this explicitly on the getter side.  Not doing this and trying to use the writer for anything other than `finish` after this will already result in a failure even without these two lines.

src/get.rs
```diff
@@ -224,8 +226,7 @@ where
                 }
 
                 // Shut down the stream
-                debug!("shutting down stream");
-                writer.shutdown().await?;
+                drop(reader);
```
flub, line 229: This is also optional, it will be dropped implicitly very soon otherwise.  It does also hide a call to `RecvStream::stop` which doesn't matter much in any case but does tell the sender that it can't use the stream anymore.  This matches the protocol.

src/protocol.rs
```diff
@@ -212,6 +213,67 @@ impl FromStr for AuthToken {
...
+/// Reasons to close connections or stop streams.
+///
+/// A QUIC **connection** can be *closed* and a **stream** can request the other side to
+/// *stop* sending data.  Both these associate and `error_code` with that operation, closing
+/// also adds a `reason` as some arbitrary bytes.
+///
+/// This enum exists so we have a single namespace for `error_code`s used.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[repr(u16)]
+pub(crate) enum Closed {
```
flub, line 225: I'm open to bikeshedding this name... It's a bit odd as it also applies to `stop` not just `close`.  But `ErrorCode` seems also odd somehow.

src/protocol.rs
```diff
@@ -212,6 +213,67 @@ impl FromStr for AuthToken {
...
+        VarInt::from(source as u16)
+    }
+}
+
+/// Unknown error_code, can not be converted into [`Closed`].
+#[derive(thiserror::Error, Debug)]
+#[error("Unknown error_code: {0}")]
+pub(crate) struct UnknownErrorCode(u64);
+
+impl TryFrom<VarInt> for Closed {
```
flub, line 264: The typesystem doesn't tell me when I missed a variant here.  That makes me a little sad.

src/provider.rs
```diff
@@ -359,90 +362,80 @@ async fn handle_stream(
         bail!("no valid handshake received");
     }
 
-    // 2. Decode protocol messages.
-    loop {
```
flub, line 363: Most of this code is simply dedented because this loop was removed and there's now a single `if let`.  I'll explicitly call out all the changes to make this more reviewable.

src/provider.rs
```diff
@@ -359,90 +362,80 @@ async fn handle_stream(
...
-                debug!("got request({})", request.id);
-                let _ = events.send(Event::RequestReceived {
-                    connection_id,
-                    request_id: request.id,
-                    hash,
-                });
+    // 2. Decode the request.
+    debug!("reading request");
+    let request = read_lp::<_, Request>(&mut reader, &mut in_buffer).await?;
+    reader.stop(Closed::RequestReceived.into())?;
```
flub, line 368: This is new, we don't allow the client to still write on this stream after we read the request.  The client has a matching finish call on this.

src/provider.rs
```diff
@@ -359,90 +362,80 @@ async fn handle_stream(
...
+                    let (status, writer1) =
+                        send_blob(db.clone(), blob.hash, writer, &mut out_buffer, request.id)
+                            .await?;
+                    writer = writer1;
+                    if SentStatus::NotFound == status {
+                        break;
                     }
                 }
 
+                writer.finish().await?;
```
flub, line 421: This is new and the source of all my troubles in this PR.  But also the raison d'etre.

We need to be sure the client has really received all the data before emitting the `TransferCompleted`.  Otherwise a user of Provider might call `shutdown` before we have really finished transferring all the data and upon closing the connection data is lost.  That's pretty flaky to track down :grin:

src/provider.rs
```diff
@@ -359,90 +362,80 @@ async fn handle_stream(
...
+                    request_id: request.id,
+                });
                 debug!("finished response");
             }
-            None => {
-                break;
+            _ => {
+                debug!("not found {}", hash);
+                write_response(&mut writer, &mut out_buffer, request.id, Res::NotFound).await?;
+                writer.finish().await?;
```
flub, line 431: This finish is also new, mirror of `TransferCompleted` case.

flub:
Third time is a charm.  Please review carefully.

src/protocol.rs
```diff
@@ -212,6 +213,67 @@ impl FromStr for AuthToken {
     }
 }
 
+/// Reasons to close connections or stop streams.
+///
+/// A QUIC **connection** can be *closed* and a **stream** can request the other side to
+/// *stop* sending data.  Both these associate and `error_code` with that operation, closing
```
ramfox: I don't understand "Both these associate and `error_code` with that operation". I'm not sure what "both these" and "that" are referring to.

src/protocol.rs
ramfox: I don't think the name `Closed` is confusing. I read it as "this is the reason the stream is closed", rather than "this has to do with the action of 'closing' rather than 'stopping' specifically".

ramfox (changes requested):
Cool, so we are now ensuring that the transfer is finished (all data is read by the `get` side) before notifying the `Event` stream that the transfer has been completed. 🔥 

This is not utilized by our CLI yet, so to "see it in action" you have to run the tests.

I only had one bit I was confused by, and it's in the `Closed` documentation!

src/get.rs
dignifiedquire: seems like good cleanup to have this

dignifiedquire: Wait actually this seems wrong, we should only close the writer once we have received the full response. While right now we don't sent acks or errors, I would expect a next extension, to actually send a final thing of "all good" or "error: xyz" when the transfer completed.

src/lib.rs
```diff
@@ -204,7 +209,13 @@ mod tests {
         let events_task = tokio::task::spawn(async move {
             let mut events = Vec::new();
             while let Ok(event) = provider_events.recv().await {
-                events.push(event);
+                match event {
+                    Event::TransferCompleted { .. } => {
```
dignifiedquire, line 213: why didn't wee need the special casing before?

src/lib.rs
```diff
@@ -204,7 +209,13 @@ mod tests {
...
+                        events.push(event);
+                        break;
+                    }
+                    _ => events.push(event),
```
dignifiedquire, line 217: should handle aborting as well, otherwise it will hang in a failure case

src/provider.rs
dignifiedquire: as mentionend on the reader side, this should only happen once the transfer is done imo

src/get.rs
flub: I think I'm slightly favour of keeping this in for now, it is how the current protocol works.  If we change the protocol then it's easy enough to change this.

(With the current protocol I don't immediately see any benefit to adding explicit ack or err responses.  There is nothing the server can do about it and it adds an extra roundtrip.  The provider already gets an ack by leveraging QUIC's built-in acks via `.finish()`.)

src/provider.rs
flub: On this side I care a little more with the current protocol.  If we don't a misbehaving client can fill up buffers on our side which is wasteful and can probably be abused for DDOS things etc.

src/lib.rs
flub: Before we shut down the provider as soon as the `get` function returned.  If this happened while the provider was suspended in the await of the new `SendStream::finish()` call then that ended up closing the connection (which is still not graceful) which meant the finish call would return in an error (`LocallyClosed`) and the `TransferCompleted` event was never sent.  Further down the test would fail due to the missing `TransferCompleted` event.

src/lib.rs
flub: I protected that with a timeout on line 260 which is needed even if `TransferAborted` is handled here.  Aborting is not the expected case so I don't think it's worth adding more code to the test here for it, given the timeout already makes the test robust.

Though you could argue that's two bits of code which are far away from each other with too tight coupling.

src/protocol.rs
flub: `s/and/an/` which made this more confusing than it needed to be.  I've tried to improve the wording.

ramfox (approved):

PR #125, opened 2023-02-08
include bao from n0, branch sendme

rklaehn:
Use the AsyncSliceDecoder in n0/bao that hopefully at some point will be merged into bao itself

This allows removing our own slice decoder.


rklaehn:
We now have our own bao, abao, and are using that since #144

PR #126, opened 2023-02-08
refactor: make on_collection take a reference

rklaehn:
This way we avoid a clone

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -116,8 +116,8 @@ mod tests {
                 hash,
                 token,
                 opts,
-                || async move { Ok(()) },
-                |_collection| async move { Ok(()) },
+                || async { Ok(()) },
+                |_collection| async { Ok(()) },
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -116,8 +116,8 @@ mod tests {
                 hash,
                 token,
                 opts,
-                || async move { Ok(()) },
-                |_collection| async move { Ok(()) },
+                || async { Ok(()) },
+                |_collection| async { Ok(()) },
```
flub, line 120: i don't quite follow why this move disappeared.  i guess it's not like it was doing anything useful :shrug:

flub (approved):

src/lib.rs
rklaehn: async move { some constant expression } just seems weird. I was thinking about replacing it with `futures::future::ok(())`, but then thought - just remove the moves.

Probably there will be a clippy lint for "useless move" at some point...

PR #129, opened 2023-02-09
Add read_total_size fn

rklaehn:
It is called read_total_size since it actually advances the internal state.


rklaehn:
Was a misunderstanding

PR #137, opened 2023-02-13
deps: Remove duplicate tokio dependency

flub:
The tokio feature of tracing-futures seems crazy outdated and not
needed.  Enabling it pulls in an ancient version of tokio for no
reason whatsoever.  Get rid of it.


rklaehn (approved):
nice!

dignifiedquire (approved):

PR #138, opened 2023-02-13
refactor: use multi threading when computing outboards

rklaehn:
I did not do this previously because something wasn't ready. But now can't remember anymore why I didn't do it...

Obviously helps mostly for large files...

### DIFF
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -556,6 +556,8 @@ impl From<&std::path::Path> for DataSource {
 ///
 /// If the size of the file is changed while this is running, an error will be
 /// returned.
+///
+/// path and name are returned with the result to provide context
 fn compute_outboard(
     path: PathBuf,
     name: Option<String>,
@@ -600,9 +602,9 @@ pub async fn create_collection(data_sources: Vec<DataSource>) -> Result<(Databas
     let mut db = HashMap::with_capacity(data_sources.len() + 1);
     let mut blobs = Vec::with_capacity(data_sources.len());
     let mut total_blobs_size: u64 = 0;
-
     let mut blobs_encoded_size_estimate = 0;
 
+    // compute outboards in parallel, using tokio's blocking thread pool
     let outboards = data_sources.into_iter().map(|data| {
         let (path, name) = match data {
             DataSource::File(path) => (path, None),
@@ -610,10 +612,13 @@ pub async fn create_collection(data_sources: Vec<DataSource>) -> Result<(Databas
         };
         tokio::task::spawn_blocking(move || compute_outboard(path, name))
     });
+    // wait for completion and collect results
     let outboards = future::join_all(outboards)
         .await
         .into_iter()
         .collect::<Result<Result<Vec<_>, _>, _>>()??;
+    // insert outboards into the database and build collection
+
     for (path, name, hash, outboard) in outboards {
         debug_assert!(outboard.len() >= 8, "outboard must at least contain size");
         let size = u64::from_le_bytes(outboard[..8].try_into().unwrap());
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -20,6 +20,7 @@ use std::{collections::HashMap, sync::Arc};
 use anyhow::{bail, ensure, Context, Result};
 use bao::encode::SliceExtractor;
 use bytes::{Bytes, BytesMut};
+use futures::future;
 use serde::{Deserialize, Serialize};
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 use tokio::sync::broadcast;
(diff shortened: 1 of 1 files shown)


rklaehn:
Speed comparison. Not very accurate since I just press Control-C as soon as I see the hash, but you get the idea...

main:
```
sendme on  main [?] is 📦 v0.1.0 via 🦀 v1.67.0 took 13s 
❯ time cargo run --release -- provide ../testdir_large_files/
    Finished release [optimized] target(s) in 0.13s
     Running `target/release/sendme provide ../testdir_large_files/`
Reading ../testdir_large_files/
Collection: bafkr4if7d6yhuz7xhmhjsd2z6v5sbgun5b6ciirimzwyazplzj3bwmlchq

- ../testdir_large_files/file1: 10 bytes
- ../testdir_large_files/test5: 1000000000 bytes

PeerID: TFNNvrnLXWflvoLultHM3QTG0V8oFs63Tw4HzbRW0Rc
Auth token: FoatKFDw3tT3d7gVMJfQ-y2qVS5J5QPMnjaHd5hL0oI
All-in-one ticket: IL8fsHpn9zsOmQ9Z9Xsgmo3ofCQiKGZtgGXrynYbMWI8IExTTb65y11n5b6C7pbRzN0ExtFfKBbOt08OB820VtEXAH8AAAHRIhaGrShQ8N7U93e4FTCX0PstqlUuSeUDzJ42h3eYS9KC
^C
Shutting down provider...
cargo run --release -- provide ../testdir_large_files/  6.53s user 0.74s system 90% cpu 8.011 total
```

this branch:
```
sendme on  multi-threaded-add [?] is 📦 v0.1.0 via 🦀 v1.67.0 took 3s 
❯ time cargo run --release -- provide ../testdir_large_files/
    Finished release [optimized] target(s) in 0.13s
     Running `target/release/sendme provide ../testdir_large_files/`
Reading ../testdir_large_files/
Collection: bafkr4if7d6yhuz7xhmhjsd2z6v5sbgun5b6ciirimzwyazplzj3bwmlchq

- ../testdir_large_files/file1: 10 bytes
- ../testdir_large_files/test5: 1000000000 bytes

PeerID: 0jSyRVTHtid2jvm-4B7rgKn3bt2Ol3fu4-TIkwTHp-8
Auth token: ABTVQeObBhXWZ0egjfcPqe3Xa6exLlitFWN0MhC1EHw
All-in-one ticket: IL8fsHpn9zsOmQ9Z9Xsgmo3ofCQiKGZtgGXrynYbMWI8INI0skVUx7Yndo75vuAe64Cp927djpd37uPkyJMEx6fvAH8AAAHRIgAU1UHjmwYV1mdHoI33D6nt12unsS5YrRVjdDIQtRB8
^C
Shutting down provider...
cargo run --release -- provide ../testdir_large_files/  6.87s user 1.12s system 350% cpu 2.280 total
```

dignifiedquire (approved):

src/provider.rs
```diff
@@ -591,25 +602,24 @@ pub async fn create_collection(data_sources: Vec<DataSource>) -> Result<(Databas
...
-            path.display()
-        );
-        // spawn a blocking task for computing the hash and outboard.
-        // pretty sure this is best to remain sync even once bao is async.
-        let path2 = path.clone();
-        let (hash, outboard) =
-            tokio::task::spawn_blocking(move || compute_outboard(path2)).await??;
-
+        tokio::task::spawn_blocking(move || compute_outboard(path, name))
+    });
```
flub, line 614: I think it might make sense to limit this concurrency somewhat.  Right now if there are 2000 files in the collection this will spawn 2000 threads I think.  Then those threads will also stay alive for an additional 10s (the default of `thread_keep_alive`.

You can have a super simple threadpool by having a single task read tasks from an mpsc together with using a Semaphore and spawning the work using `spawn_blocking`.  Much like this: https://github.com/getsentry/relay/blob/1e470ec87aee8fb9214844a1cf5eaa5257ea3d69/relay-server/src/actors/processor.rs#L2486-L2496

flub: so default of `max_blocking_threads` is 512, so we'll only spawn a quarter of those 2000 threads.  still, i think this is worth it

rklaehn: Weird. You would think that you just tell tokio what to do, and it will take care of doing it.

I guess that spawn_blocking is more meant to be used for blocking io tasks and not for cpu bound tasks. But the task we want to speed up are in fact _mostly_ IO tasks - bao outboard computation is very fast, so this is dominated by IO.

So if it is not good to use spawn_blocking to spawn a very large number of IO tasks - why then does it allow you to do so?

rklaehn: Now, I could just go ahead and use rayon. And before I write my own threadpool I would prefer to do that. But it seems weird to use rayon for this task that is mostly io...

flub: Oh, all those tasks are IO-bound and not CPU bound?  That changes the picture entirely and maybe it is fine to leave as is then.  The only other reason to limit the concurrency of this would be if sendme needs to have capacity to do other `spawn_blocking` work done at the same time of this computation.  But I guess that's not really the case.

flub: > Now, I could just go ahead and use rayon. And before I write my own threadpool I would prefer to do that. But it seems weird to use rayon for this task that is mostly io...

If you need nothing from rayon's computation utilities combining spawn_blocking with a semaphore is a much better choice then using rayon I think.  There's nothing wrong for using it to run some cpu-bound work.  Only need to make sure you're not exhausting the other usages of the spawn_blocking threadpool.

flub (approved):

PR #139, opened 2023-02-13
fix: fail if an on_blob function does not read all the content

### LINKED ISSUE #127: Misbehaving on_blob callback breaks entire transport
The `get::run` function relies on the `on_blob` callback to fully consume the reader it receives.

This could be made more robust, e.g. the `get::run` method could call `tokio::io::copy(&mut blob_reader, &mut tokio::io::sink()).await?` right after the callback returns.

Another much more invasive option is to change the API so that the callback is supposed to return an `AsyncWrite` thing.  Probably not worth it.

A prototype test for this is: https://gist.github.com/flub/48325fb36476b7caf4d8e8e733d02a4f though it does not fail because the files are too small, making them larger should make it fail.

### LINKED ISSUE #718: Misbehaving on_blob callback breaks entire transport

ramfox:
closes n0-computer/iroh#718

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -331,7 +331,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_blob_reader_partial() {
+    async fn test_blob_reader_partial() -> Result<()> {
         // Prepare a Provider transferring a file.
         let dir = testdir!();
         let src0 = dir.join("src0");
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -253,4 +253,54 @@ mod tests {
...
+        // Prepare a Provider transferring a file.
+        let dir = testdir!();
+        let src0 = dir.join("src0");
+        let src1 = dir.join("src1");
+        let src0_data = {
+            let mut content = vec![0u8; 10000];
+            rand::thread_rng().fill_bytes(&mut content);
+            content
+        };
+        fs::write(&src0, &src0_data).await?;
```
flub: maybe put this in a nested block so those 10000 bytes are dropped as soon as they're written to the file?  bonus points for never storing them all in memory, but while a good exercise it's only a test...

src/lib.rs
```diff
@@ -253,4 +253,54 @@ mod tests {
...
+        fs::write(&src1, "hello world").await?;
+        let (db, hash) = create_collection(vec![src0.into(), src1.into()]).await?;
+        let provider = Provider::builder(db)
+            .bind_addr("127.0.0.1:0".parse().unwrap())
+            .spawn()?;
+        let auth_token = provider.auth_token();
+        let provider_addr = provider.listen_addr();
+
+        let timeout = tokio::time::timeout(
+            std::time::Duration::from_millis(300),
```
flub: Because CI is the flaky stuff it tends to be I usually pick stupid numbers in the range of tens of seconds.  The delay only kicks in in case of test failure anyway.

flub (approved):
nice!

src/lib.rs
ramfox: ahh, yea, didn't put together that since it only happens on failure, there is no reason to try to optimize this for as low as possible. thanks!

PR #141, opened 2023-02-13
feat: Add generic progress emitter utility

flub:
This adds a generic utility which can be used to track progress of
data transfers by emitting progress via a tokio broadcast channel.

This is extracted from the deltachat PR:
https://github.com/deltachat/deltachat-core-rust/pull/4007

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -1,0 +1,143 @@
+//! Generic utilities to track progress of data transfers.
+//!
+//! This is not especially specific to sendme but can be helpful together with it.  The
+//! [`ProgressEmitter`] has a [`ProgressEmitter::wrap_async_read`] method which can make it
+//! easy to track process of transfers.
+//!
+//! However based on your environment there might also be better choices for this, e.g. very
+//! similar and more advanced functionality is available in the `indicatif` crate for
+//! terminal applications.
+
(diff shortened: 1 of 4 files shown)


src/progress.rs
```diff
@@ -0,0 +1,143 @@
...
+    inner: Arc<InnerProgressEmitter>,
+}
+
+impl ProgressEmitter {
+    /// Creates a new emitter.
+    ///
+    /// The emitter expects to see *total* being added via [`ProgressEmitter::inc`] and will
+    /// emit *steps* updates.
+    pub fn new(total: u64, steps: u16) -> Self {
+        let (tx, _rx) = broadcast::channel(16);
```
dignifiedquire, line 40: should this 16 be configurable?

dignifiedquire (approved):
nice

src/progress.rs
flub: My opiniated answer is: no.  At least not until someone has a good usecase.

My reasoning:

Firstly, you should be able to keep up: spawn a task.

Secondly it is only a progress report.  If you happen to hit the `RecvErr::Lagged` branch you still have access to the most up to date progress.  The historical progress updates are not really useful anymore, there is not event a timestamp associated with them.

PR #143, opened 2023-02-14
Fix QUIC stream completion

flub:
Sending a STOP_STREAM frame from the receiver is the way for a receiver to go to the reset state.  It involves sending an extra frame, the receiver will then send its own RESET_STREAM frame and the stream is considered ended in an error state.

Instead the right way to end a stream is to set the FIN bit on the final STREAM frame, ideally not even resulting in an extra frame or packet sent.  This is done on the sender side.

### DIFF
--- a/src/get.rs
+++ b/src/get.rs
@@ -158,7 +158,6 @@ where
         let used = postcard::to_slice(&req, &mut out_buffer)?;
         write_lp(&mut writer, used).await?;
     }
-    dbg!(writer.finish().await)?;
     drop(writer);
 
     // 3. Read response
--- a/src/get.rs
+++ b/src/get.rs
@@ -158,6 +158,7 @@ where
         let used = postcard::to_slice(&req, &mut out_buffer)?;
         write_lp(&mut writer, used).await?;
     }
+    writer.finish().await?;
     drop(writer);
 
     // 3. Read response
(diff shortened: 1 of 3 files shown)


ramfox (approved):
Okay, so to summarize what I've learned here:
- the graceful way to ensure the stream is closed is from the **sender** side, by using `SendStream::finish`
- We are already doing this by calling `writer.finish()` in `get::run`
- using `RecvStream::stop` is how the receiving side "resets" a stream, and it involves a back and forth: the receiving side sends `STOP_STREAM` & the sending side sends `RESET_STREAM` back. This also means the stream is considered to have ended in an error state.
- so no need to "mirror" the action on the provider side by calling `reader.stop()`

dignifiedquire (approved):

flub:
> Okay, so to summarize what I've learned here:
> 
>     * the graceful way to ensure the stream is closed is from the **sender** side, by using `SendStream::finish`
> 
>     * We are already doing this by calling `writer.finish()` in `get::run`
> 
>     * using `RecvStream::stop` is how the receiving side "resets" a stream, and it involves a back and forth: the receiving side sends `STOP_STREAM` & the sending side sends `RESET_STREAM` back. This also means the stream is considered to have ended in an error state.
> 
>     * so no need to "mirror" the action on the provider side by calling `reader.stop()`

Indeed, at least that's my current understanding.  I currently think the "right" way to handle this on the receiving side is to listen for any new data and only if new data arrives it should send the stop frame.  But that's a bit overkill.

PR #144, opened 2023-02-15
ref: use our own bao crate abao

rklaehn:
Requires just a tiny code change since bao takes a reference of a hash instead of the value.


dignifiedquire (approved):
very nice

PR #148, opened 2023-02-16
start debugging earyl close

dignifiedquire:
When finished with a stream you have to read to EOF.  Sometimes the sender could be sending the FIN bit in an empty STREAM frame after all the data is sent, if the receiver did not read EOF in this case the receiver will still implicitly call `RecvStream::stop` on drop, making the stream error for the sender.

----

running a couple times: should reproduce the issue
```
> RUST_LOG=debug cargo test server_close -- --nocapture
```

### DIFF
--- a/src/get.rs
+++ b/src/get.rs
@@ -229,6 +229,12 @@ where
                     }
                 }
 
+                let chunk = reader.read_chunk(8, false).await?;
+                ensure!(
+                    chunk.is_none(),
+                    "received unexpected data from the provider: {:?}",
+                    chunk
+                );
                 // Shut down the stream
                 drop(reader);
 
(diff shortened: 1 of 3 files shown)


src/get.rs
```diff
@@ -229,6 +229,12 @@ where
                     }
                 }
 
+                let chunk = reader.read_chunk(8, false).await?;
+                ensure!(
+                    chunk.is_none(),
+                    "received unexpected data from the provider: {:?}",
+                    chunk
+                );
```
flub: I think at this point we should call `reader.stop()` to send back an error to the sender

src/get.rs
```diff
@@ -230,6 +230,10 @@ where
                 }
 
                 // Shut down the stream
+                if let Some(chunk) = reader.read_chunk(8, false).await? {
+                    reader.stop(0u8.into()).ok();
+                    error!("Received unexpected data from the provider: {chunk:?}");
```
flub, line 235: I'm choosing not to error the get operation itself here, because it knows for a fact that all the blobs were received and written entirely correctly.

PR #150, opened 2023-02-17
fix: ensure we emit `TransferAborted` event if anything goes wrong during the transfer

### LINKED ISSUE #146: Ensure TransferAborted send when connection drops
We seem to have some protocol flakyness still that I'm hunting down.  This caused this error once in CI: https://github.com/n0-computer/sendme/actions/runs/4184967389/jobs/7251348335
Basically the test times out waiting for the provider to emit the event that it's done.

This is almost certainly caused due to this unresolved todo: https://github.com/n0-computer/sendme/blob/main/src/provider.rs#L432

So let's fix that.

### LINKED ISSUE #713: Ensure TransferAborted send when connection drops

ramfox:
potentially closes n0-computer/iroh#713 

This refactor splits the `handle_stream` function up into smaller (logical) functions.  If these functions error, we can emit a `TransferAborted` event before returning the error.

This has the added benefit of making the `handle_stream` function a bit easier to take in. None of these new functions have any code changes, the only difference is how errors are handled and when we return.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -215,6 +215,10 @@ mod tests {
                         events.push(event);
                         break;
                     }
+                    Event::TransferAborted { .. } => {
+                        events.push(event);
+                        break;
+                    }
                     _ => events.push(event),
                 }
             }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -211,11 +211,7 @@ mod tests {
             let mut events = Vec::new();
             while let Ok(event) = provider_events.recv().await {
                 match event {
-                    Event::TransferCompleted { .. } => {
-                        events.push(event);
-                        break;
-                    }
-                    Event::TransferAborted { .. } => {
+                    Event::TransferCompleted { .. } | Event::TransferAborted { .. } => {
                         events.push(event);
                         break;
                     }
@@ -269,9 +265,16 @@ mod tests {
         provider.shutdown();
         provider.await?;
 
-        assert_eq!(events.len(), 3);
+        assert_events(events);
 
         Ok(())
+    }
+
+    fn assert_events(events: Vec<Event>) {
+        assert_eq!(events.len(), 3);
+        assert!(matches!(events[0], Event::ClientConnected { .. }));
+        assert!(matches!(events[1], Event::RequestReceived { .. }));
+        assert!(matches!(events[2], Event::TransferCompleted { .. }));
     }
 
     fn setup_logging() {
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -438,6 +438,7 @@ async fn transfer_collection(
             send_blob(db.clone(), blob.hash, writer, buffer, request_id).await?;
         writer = writer1;
         if SentStatus::NotFound == status {
+            write_response(&mut writer, buffer, request_id, Res::NotFound).await?;
             writer.finish().await?;
             return Ok(status);
         }
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -232,8 +232,9 @@ pub enum Event {
     TransferAborted {
         /// The quic connection id.
         connection_id: u64,
-        /// The request id.
-        request_id: u64,
+        /// The request id. When `None`, the transfer was aborted before or during reading and decoding
+        /// the transfer request.
+        request_id: Option<u64>,
     },
 }
 
@@ -338,7 +339,12 @@ async fn handle_connection(
     .await
 }
 
-async fn handshake<R: AsyncRead + Unpin>(
+/// Read and decode the handshake. Will fail if there is an error while reading, there is a token
+/// mismatch, or no valid handshake was received.
+///
+/// When successful, the reader is still useable after this function and the buffer will be drained of any handshake
+/// data.
+async fn read_handshake<R: AsyncRead + Unpin>(
     mut reader: R,
     buffer: &mut BytesMut,
     token: AuthToken,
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -215,6 +215,10 @@ mod tests {
                         events.push(event);
                         break;
                     }
+                    Event::TransferAborted { .. } => {
+                        events.push(event);
+                        break;
```
ramfox, line 216: We should also break when we get a `TransferAborted` event.

src/provider.rs
```diff
@@ -338,111 +338,165 @@ async fn handle_connection(
...
+    (mut writer, mut reader): (quinn::SendStream, quinn::RecvStream),
+    events: broadcast::Sender<Event>,
+) -> Result<()> {
+    let mut out_buffer = BytesMut::with_capacity(1024);
+    let mut in_buffer = BytesMut::with_capacity(1024);
+
+    // 1. Read Handshake
+    debug!("reading handshake");
+    if let Err(e) = handshake(&mut reader, &mut in_buffer, token).await {
+        notify_transfer_aborted(events, connection_id, 0);
```
ramfox: What should we put for the `request_id` (in the `TransferAborted` enum), before we get a request?

I settled on `0`, but would love to know if there are better ideas.

src/provider.rs
```diff
@@ -338,111 +338,165 @@ async fn handle_connection(
...
+        return Err(e);
+    }
+
+    // 2. Decode the request.
+    debug!("reading request");
+    let request = match decode_request(reader, &mut in_buffer).await {
+        Ok(r) => r,
+        Err(e) => {
+            notify_transfer_aborted(events, connection_id, 0);
```
ramfox: same question as above

src/provider.rs
```diff
@@ -338,111 +338,166 @@ async fn handle_connection(
...
+    let (outboard, data) = match db.get(&hash) {
+        // We only respond to requests for collections, not individual blobs
+        Some(BlobOrCollection::Collection(d)) => d,
+        _ => {
+            debug!("not found {}", hash);
+            notify_transfer_aborted(events, connection_id, request.id);
+            write_response(&mut writer, &mut out_buffer, request.id, Res::NotFound).await?;
+            writer.finish().await?;
+
+            return Ok(());
```
ramfox, line 515: The only case where we emit `TransferAborted`, but the entire process does not end in error is when the hash is `NotFound`. 

Are there any other cases where we should be ending without an error? I think "no", but wanted to bring it up.

src/lib.rs
flub: Yes, though you can merge this with the `Event::TransferCompleted { ... } | Event::TransferAborted { .. } => `

src/provider.rs
```diff
@@ -338,111 +338,166 @@ async fn handle_connection(
     .await
 }
 
-async fn handle_stream(
-    db: Database,
+async fn handshake<R: AsyncRead + Unpin>(
```
flub: maybe call it `read_handshake` since it consumes from the stream?

I'm also a big fan of doc comments everywhere :)  They can say useful things like what state `reader` and `buffer` are in after having called this.  When is this supposed to fail?

src/provider.rs
```diff
@@ -338,111 +338,166 @@ async fn handle_connection(
...
     } else {
         bail!("no valid handshake received");
     }
+    Ok(())
+}
 
-    // 2. Decode the request.
-    debug!("reading request");
-    let request = read_lp::<_, Request>(&mut reader, &mut in_buffer).await?;
+async fn decode_request(mut reader: quinn::RecvStream, buffer: &mut BytesMut) -> Result<Request> {
```
flub: nit, more like `read_request`.  "decode" usually implies you already have the encoded thing.

src/provider.rs
```diff
@@ -338,111 +338,166 @@ async fn handle_connection(
...
-                    let (status, writer1) =
-                        send_blob(db.clone(), blob.hash, writer, &mut out_buffer, request.id)
-                            .await?;
-                    writer = writer1;
-                    if SentStatus::NotFound == status {
-                        break;
-                    }
-                }
+    writer.finish().await?;
+    debug!("finished response");
```
flub: all the other log messages are in the `handle_stream` function...

src/provider.rs
flub: Make `request_id` an `Option` in `Event::TransferAborted`?  That also gives you a place to document why it can be `None`.

src/provider.rs
flub: i think this is fine

src/provider.rs
```diff
@@ -338,111 +338,166 @@ async fn handle_connection(
...
+            write_response(&mut writer, &mut out_buffer, request.id, Res::NotFound).await?;
+            writer.finish().await?;
+
+            return Ok(());
         }
+    };
+
+    // 5. Transfer data!
+    if let Err(e) = transfer_collection(
+        db.clone(),
```
flub: Not that this matters since clones are cheap.  But you can probably pass the database as a reference since you await it right away.  You pass `out_buff` as reference anyway as well

flub (approved):

src/lib.rs
ramfox: ahhh, yes, sorry I forgot to ask the question I wanted to ask.

For the `Event::TransferAborted`, I can't tell if it would be helpful to have the `event_task` emit an error. Or if it's enough to just rely on the tests to otherwise fail if the transfer doesn't happen correctly.

flub: Hum, I think the test only checks for the number of events.  But currently this task doesn't return a Result which is kind of nice and simple.  Maybe just change the assert statement that checks the len into checking the events in order expected might be the simplest way to improve this?

src/provider.rs
ramfox: good point

src/lib.rs
ramfox: ahh, that's a great & much more simple idea, thank you.

PR #151, opened 2023-02-17
Daemonize sendme

rklaehn:
implements n0-computer/iroh#720

This currently uses quic-rpc with the quinn transport on a separate port. We can change this to use an UDS, but given that what flows over the RPC are small commands, it won't be an issue WRT performance.

rpc-protocol contains the protocol definition. Defining how a message behaves might read a bit verbose, and could be made much shorter using existing macros. However, I don't mind the verbosity that much.

The Version/Watch stuff is shamelessly copied from iroh.

For now this keeps the old provide command in place. Not sure we want to completely remove the provide_interactive code path, since I think @Arqu uses it in his tests, and also a fair bit of work went into it.

This does comply with our 1.63 MSRV. I had to publish an entirely new quic-rpc to rip out GATs so we can do that.


Arqu:
> For now this keeps the old provide command in place. Not sure we want to completely remove the provide_interactive code path, since I think @Arqu uses it in his tests, and also a fair bit of work went into it.

`start` is basically the same as `provide` right? In that case feel free to kill the provide, it's a very minor adaptation on my end and would rather drop cruft for the cost of incremental fixes to my scripts.

rklaehn:
> `start` is basically the same as `provide` right? In that case feel free to kill the provide, it's a very minor adaptation on my end and would rather drop cruft for the cost of incremental fixes to my scripts.

No, start just starts the service without providing something. You then have to connect via rpc to _actually_ provide something. Maybe just have the option to add initial data when calling start?

Arqu:
In that case I do have a bit more work. Can we then keep it and I'll try to amend that next week so we can remove.

src/main.rs
```diff
@@ -1,19 +1,29 @@
...
+    ListRequest, ProvideRequest, SendmeRequest, SendmeResponse, SendmeService, VersionRequest,
+};
+use sendme::{protocol::AuthToken, provider::Database};
 use tokio::io::AsyncWriteExt;
 use tokio::sync::Mutex;
 use tracing_subscriber::{prelude::*, EnvFilter};
 
 use sendme::{get, provider, Hash, Keypair, PeerId};
 
+const RPC_PORT: u16 = 0x1337;
```
flub, line 25: lol, probably a bit too busy a port

src/main.rs
```diff
@@ -40,6 +50,26 @@ enum Commands {
...
+        addr: Option<SocketAddr>,
+        /// Auth token, defaults to random generated.
+        #[clap(long)]
+        auth_token: Option<String>,
+        /// If this path is provided and it exists, the private key is read from this file and used, if it does not exist the private key will be persisted to this location.
+        #[clap(long)]
+        key: Option<PathBuf>,
+    },
+
+    /// Start sendme as a service
```
flub, line 67: wrong doc comment?

src/main.rs
```diff
@@ -40,6 +50,26 @@ enum Commands {
         #[clap(long)]
         key: Option<PathBuf>,
     },
+    /// Start sendme as a service
+    #[clap(about = "Start sendme as a service")]
+    Start {
```
flub, line 55: Would this not work as a single new command line flag to `provide` which enables listening to the rpc addr instead of a new command?  (ok, maybe two: `--rpc` & `--rpc-bind-addr=xxx`)

src/provider.rs
```diff
@@ -34,30 +40,49 @@ use crate::blobs::{Blob, Collection};
...
 
 const MAX_CONNECTIONS: u32 = 1024;
 const MAX_STREAMS: u64 = 10;
+const HEALTH_POLL_WAIT: Duration = Duration::from_secs(1);
 
 /// Database containing content-addressed data (blobs or collections).
-#[derive(Debug, Clone)]
-pub struct Database(Arc<HashMap<Hash, BlobOrCollection>>);
+#[derive(Debug, Clone, Default)]
+pub struct Database(Arc<Mutex<HashMap<Hash, BlobOrCollection>>>);
```
flub, line 56: I wonder if an RwLock is worth it?  That assumes there are many concurrent gets and only the occasional add.

src/provider.rs
```diff
@@ -34,30 +40,49 @@ use crate::blobs::{Blob, Collection};
...
 
 impl Database {
-    fn get(&self, key: &Hash) -> Option<&BlobOrCollection> {
-        self.0.get(key)
+    fn get(&self, key: &Hash) -> Option<BlobOrCollection> {
+        self.0.lock().unwrap().get(key).cloned()
+    }
+
+    fn union_with(&self, db: HashMap<Hash, BlobOrCollection>) {
```
flub, line 63: bikeshedding, I would expect a name like this to not modify anything and return a new Db instead.  This I would call something like `merge` or `add` or `merge_db` or ...

src/rpc_protocol.rs
```diff
@@ -0,0 +1,97 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ListResponse {
+    pub path: PathBuf,
+    pub hash: Hash,
+    pub size: u64,
+}
+
+impl Msg<SendmeService> for ListRequest {
+    type Pattern = ServerStreaming;
+}
```
flub, line 38: It's a bit odd for the casual reader that RpcMsg doesn't need this part.

src/tls.rs
```diff
@@ -20,7 +20,7 @@ use ssh_key::LineEnding;
 use crate::util;
 
 // TODO: change?
-const P2P_ALPN: [u8; 6] = *b"libp2p";
+pub(crate) const P2P_ALPN: [u8; 6] = *b"libp2p";
```
flub, line 23: we really should stop borrowing someone else's ALPN.  That's just bad.  Let's change it to `n0/sendme/1` or something, we can then still change it later.

src/main.rs
```diff
@@ -186,6 +216,20 @@ impl FromStr for Blake3Cid {
     }
 }
 
+async fn make_rpc_client(
+) -> anyhow::Result<RpcClient<SendmeService, QuinnConnection<SendmeResponse, SendmeRequest>>> {
+    let endpoint = sendme::get::make_client_endpoint(None, vec!["rpc".as_bytes().to_vec()])?;
```
flub, line 221: Since I'm making up ALPNs, let's change this to `n0/sendme-rpc/1`

src/main.rs
rklaehn: I checked in some port list and did not find anything obvious...

src/main.rs
rklaehn: Yes, probably best. I just implemented things as specified in the issue for now, and there it says to have start.

src/provider.rs
rklaehn: Might be worth it. Note that I did not even reach for parking_lot this time. I think there have been some improvements to the stdlib lock to make it faster...

src/provider.rs
rklaehn: Not sure if there is a convention. I usually use e.g. `union` for a fn that produces a new value, and `union_with` for something that works in place, e.g. in https://docs.rs/radixdb/latest/radixdb/struct.RadixTree.html

The in place operations and the functional operations should share the op name. But the _with probably comes from scala.

Maybe union_mut? 🤷‍♂️

src/rpc_protocol.rs
rklaehn: Oh well, now we really get into bikeshedding. I made RpcMsg automatically implement Msg with the right pattern, since I assumed it would be the most popular one.

I can also remove this, but remember that I want this to be relatively low boilerplate even when not using macros.

I added a bunch of optional very simple macros experimentally, e.g.

```rust
declare_server_streaming!(SendmeService, ListRequest, ListResponse);
```

This is incredibly simple as far as macros go, but I still would like it to be possible to do no macros at all.

src/tls.rs
rklaehn: Yes.

dignifiedquire:
Please reopen under https://github.com/n0-computer/iroh

PR #153, opened 2023-02-17
feat: Add option to use SSLKEYLOGFILE env var

flub:
This adds support for the SSLKEYLOGFILE environment variable which is
a common standard used by browsers to log the TLS keys used for
connections.  It allows traffic inspectors like tcpdump and wireshark
to use the logged keys and decrypt the traffic.  Using this is a first
step to allowing us to observe our protocol on the wire.

The support is still hidden behind a commandline option so that it can
not be accidentally enabled by only having an environment variable
set.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -39,6 +39,9 @@ enum Commands {
         /// If this path is provided and it exists, the private key is read from this file and used, if it does not exist the private key will be persisted to this location.
         #[clap(long)]
         key: Option<PathBuf>,
+        /// Log SSL pre-master key to file in SSLKEYLOGFILE environment variable.
+        #[clap(long)]
+        keylog: bool,
     },
     /// Fetch some data by hash.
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -253,10 +269,11 @@ async fn main() -> Result<()> {
             addr,
             auth_token,
             key,
+            keylog,
         } => {
             tokio::select! {
                 biased;
-                res = provide_interactive(path, addr, auth_token, key) => {
+                res = provide_interactive(path, addr, auth_token, key, keylog) => {
```
dignifiedquire, line 276: this is starting to become a lot of arguments :/

dignifiedquire (approved):

src/main.rs
flub: Yeah, I also had that feeling but didn't do anything about it yet.

PR #155, opened 2023-02-20
feat: Use our own ALPN

flub:


src/tls.rs
```diff
@@ -20,7 +20,7 @@ use ssh_key::LineEnding;
 use crate::util;
 
 // TODO: change?
-const P2P_ALPN: [u8; 6] = *b"libp2p";
+const P2P_ALPN: [u8; 9] = *b"n0/iroh/1";
```
dignifiedquire, line 22: probably update the comment as well? 😅

flub: :joy:

dignifiedquire (approved):

